PR #11, opened 2015-03-05
add soql function magnitude

sokrahta:
Pair CR: @chitang

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -79,6 +79,8 @@ object SoQLFunctions {
   val UnaryPlus = Function("unary +", SpecialFunctions.Operator("+"), Map("a"->NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"))
   val UnaryMinus = Function("unary -", SpecialFunctions.Operator("-"), Map("a"->NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"))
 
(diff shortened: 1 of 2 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -79,6 +79,8 @@ object SoQLFunctions {
   val UnaryPlus = Function("unary +", SpecialFunctions.Operator("+"), Map("a"->NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"))
   val UnaryMinus = Function("unary -", SpecialFunctions.Operator("-"), Map("a"->NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"))
 
+  val Magnitude = Function("magnitude", FunctionName("magnitude"), Map("a"->NumLike, "b"->NumLike), Seq(VariableType("a"), VariableType("b")), Seq.empty, VariableType("b"))
```
znep: Should the result of this function be a SoQLNumber regardless of input type?

chitang: I think it should follow the first argument in the soql function - which is "a".  We were confused between which one is the log based.

sokrahta: The result is always a positive integer. Similarly the base argument is always a positive integer.

chitang: Yes, the result is always positive integer.  But there is value in following the datatype of the primary argument.  Postgres round function does that:

Function: round(dp or numeric)
Return Type: (same as input)
Description: round to nearest integer

sokrahta: that's fair, thanks chi

PR #14, opened 2015-04-01
Geometry SoQL improvements

velvia:
- Add a WKB64 (base64 WKB) translator
- Use thread local encoders/decoders for speed improvement
- Add a SoQLBinaryGeometry for pass-through / more efficient encoding


urmilan:
:+1:

velvia:
@chitang so OK to merge?

PR #16, opened 2015-05-15
Refs CORE-5412

rozap:
* Added polygon, line, and multipoint types
* added tests for wkt adn wkb serialization
* todo: remove the snapshot reference in dependencies


urmilan:
:shipit: Feel free to merge once you've replaced the snapshot dependency.

PR #18, opened 2015-07-07
SoQLPack library

velvia:
Overall, this PR moves SoQLPack logic from soda-fountain and tileserver to a centralized place here in soql-reference and extends support to far more types.

- Generic API to encode an `Iterator[Array[SoQLValue]]` to a binary SoQLPack stream, and back
- SoQLPackDumper utility to dump out a binary SoQLPack stream on STDIN
- Support for every SoQL type except for the JSON types (SoQLArray, SoQLObject, etc).  These will be added later, in this PR or another one.

The plan after this is to 
a) replace code in soda-fountain and tileserver to use this API
b) enable tileserver to use SoQLPack for all queries
c) enable RegionCoder to pull from soda fountain using SoQLPack

It might be easier to review the individual commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+soql-reference provides type definitions and utilities such as analyzers and parsers for Socrata's SoQL query language.
--- a/README.md
+++ b/README.md
@@ -1,1 +1,13 @@
 soql-reference provides type definitions and utilities such as analyzers and parsers for Socrata's SoQL query language.
+
+### soql-types
+
+This contains the definitions of all the types in SoQL.
+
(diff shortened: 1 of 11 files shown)


chitang:
Add comment near SoQLID that this is meant for internal usage only because SoQLID is not encrypted - lgtm up to 52c630f67b.

velvia:
@chitang addressed your CR and also added the remaining missing JSON support.   It's probably ready to merge after this.  I'll work on incorporating these changes into soda-fountain and fileserver.

PR #20, opened 2015-08-05
Push some app-level functionality into SoQL-Pack

kaidamasaki:


kaidamasaki:
@rjmac I assigned you because I thought you might appreciate the first commit I made here.

chitang:
:shipit: 

My concern with SoQLGeoRow.properties is that you are doing a match possibly on every row.  If possible, build an array of appropriate functions once that convert SoQLValue to Json.

kaidamasaki:
@chitang: Is scala's match slow enough that the optimization is worth it?  

Also, going to merge.  Though I'd definitely be up to improving the performance of this later, especially when we exercise this code path more heavily.

PR #25, opened 2015-09-22
Blob values

rjmac:
Blobs are, as far as SoQL proper is concerned, an opaque identifier
which can be compared for equality with other blobs, but are not
ordered.

There'll need to be a version bump to actually use it, of course.


AEscobarCruz:
lgtm

PR #32, opened 2015-11-20
Support clear id for OBE compatibility.

chitang:


soql-types/src/main/scala/com/socrata/soql/types/SoQLType.scala
```diff
@@ -77,6 +77,25 @@ case object SoQLID extends SoQLType("row_identifier") {
       obfuscator.obfuscate(soqlId)
   }
 
+  /**
+   * This exists for OBE compatibility reason.
```
znep: Can you describe exactly what this is here?

chitang: This is for presenting obe nbe id as obe sid.  We are not creating a new column, are we?

rjmac:
It saddens me, but I bow to necessity.  LGTM, esp if that unused parameter really does need to be there.

PR #33, opened 2015-12-01
Chained selects

rjmac:
This will let us express queries-on-queries directly in soql, allowing analysis to occur close to the data store, and also allowing queries-on-group-bys as a first-class citizen.


chrismetcalf:
@rjmac OK this I really want to see an example of, it could be pretty awesome.

rjmac:
`select region, sum(population) as total_pop group by region |> select avg(total_pop) as average_region_population`

Doesn't get us JOINs, but makes implementing views-on-views easier (also hidden columns, which just become "columns that aren't selected").

znep:
As a general question, what are the compatibility issues here and how much work/time will it be to upgrade dependencies?  Specifically, we have upcoming changes next sprint that will require changes in soql-reference and I want to make sure those don't end up blocked behind a larger work in progress.

chrismetcalf:
@rjmac And that's something I could pass in the `$query` parameter? I assume both the select and the subselect need to be on the same dataset?

Also, I kinda like the `|>` sugar, it looks cool. :smile:

rjmac:
@znep the main compatibility change is that instead of a SoQLAnalysis, you get a Seq[SoQLAnalysis] out of the analyzer.  It has an obvious naïve translation into subselects for postgresql, but if we want to move faster than that, the whole thing still supports unchained queries if you explicitly ask for them.

@chrismetcalf Yep, you could pass it in `$query`, though the intent's more for them to be generated by core.  It also has a more SQLish subselect syntax: `select avg(total_pop) as avg_reg_pop FROM (select region, sum(population) as total_pop group by region)`.  I couldn't decide which I liked better, so I did both since they have an obvious meaning when combined.

znep:
Please please please lets only have one syntax!  We already have way too
many duplicate ways to express things that cause confusion because everyone
ends up having to understand every form.

chrismetcalf:
I'm with @znep on the one syntax to rule them all. And as fun as `|>` is, `FROM` will be a more familiar convention to most people.

rjmac:
But harder to generate for in its primary use case.  You can just glue strings together without needing to parse anything to generate a query-on-a-view.  If I were to choose one, I'd choose `|>`.

chrismetcalf:
Hmm, do you mean because you need to surround the `FROM`'s subselect in parens?

znep:
But a computer is generating that use case, I'm sure we can teach a
computer how do do that fairly easily... no?

"|>" is pretty cryptic to a user.

Remember, this is effectively a major long term addition to our API not
just a migration thing so we should treat it as such.

chrismetcalf:
Yeah, still in favor of `FROM`, just trying to understand @rjmac's suggestion :)

rjmac:
No, because it goes in the middle of the query. If you're given `select a,b,c where x > 5` on a view you need to generate `select a, b, c from (select ...) where x > 5` on the underlying dataset.  Makes error reporting better too, because all errors will be at a uniform offset from the start of the query and so positions can be backfilled to match the user's input better.

chitang:
lgtm but it sounds like Chris/Marc prefer select ...  from (select ...).

chrismetcalf:
I'm leaning in the `from` direction, but then again I strangely find that a good portion of the developers that end up in my support queue actually don't know SQL, so they won't have a preconceived notion there. 

But yeah, still leaning `from`.

znep:
Yeah... the FROM syntax is much more human friendly and understandable.

chrismetcalf:
:shipit:

chrismetcalf:
Note to self, we went with the `SELECT event_clearance_group, count(*) GROUP BY event_clearance_group |> SELECT COUNT(*)` syntax. Exploring availability in production.

timothyclemansinsea:
I think it would be easier for newbies like me to remember to add just `$|>=count(*)` instead of converting `$select=event_clearance_group, count(*)&$group=event_clearance_group` to `SELECT event_clearance_group, count(*) GROUP BY event_clearance_group |> SELECT COUNT(*)`

chrismetcalf:
@rjmac @znep Can you chain multiple queries together, like `SELECT a, b |> SELECT a WHERE a > 50 |> SELECT MAX(a)`? (Not that that is the best way to do that query)

I agree it'd be nice to have an add-on SoQL parameter for subselects, but I want to think about how that'd affect chaining.

timothyclemansinsea:
@chrismetcalf what about $|>1= $|>2= for chaining?

PR #35, opened 2015-12-08
EN-24347 Add date_trunc functions for fixed_timestamp.

chitang:


znep:
I don't have any context around this so apologize if this is a silly question... what is the thought on how these handle time zones?  Are they all just implicitly converted to GMT before truncation or should we be specifying an explicit time zone to be converted to?

Is there a specific use case for these?

chitang:
date trunc for fixed timestamp is always performed in the system timezone.  For us, it is GMT.  If we need more flexibility around timezone translation, I suggest that we provide other functions.  Translation should be independent of date trunc.  You can apply them before or after.

chitang:
Currently, our obe date (with timezone) type is migrated as floating timestamp (w/o timezone).  I believe this is wrong.  I guess we made that choice because we need a truncation function that does not exist yet for fixed timestamp.

znep:
As far as I understand, migrating them that way was an explicit decision we
made based on the observation that (1) the vast majority of columns
appeared to have been created as with timezone by accident or
unintentionally and converting them to a floating timestamp based in the
customer's time zone actually was a significant improvement and (2)
converting them to GMT would result in most data lens pages not being what
the customer would expect or want...

There are also significant performance impacts of doing arbitrary time zone
conversions since you then can't serve them off rollups.

chitang:
That makes sense.  We probably still need the same date trunc functions for fixed timestamp though unless we are going to remove fixed timestamp for custom use.  Performance is another issue we need to handle separately.

chrismetcalf:
@chitang So what are the implications of this change on my developers? Are the same date trunc functions now available on both the floating and non floating versions of dates?

znep:
My concern is that for fixed dates it would implicitly and confusingly
(IMO) convert to UTC with no other option, which makes the use cases
somewhat limited and non-intuitive.

I believe what we found in the past was most fixed date columns appear to
be the result of people making the wrong choice by accident.

chrismetcalf:
Whoa is this some sort of zombie PR?

urmilan:
:shipit:

rjmac:
Honestly, truncating a fixed_timestamp doesn't even make sense.  We render it into UTC on the wire for developer convenience, and should be rendering it in the user's local timezone for UI purposes on the client, but in the absence of a particular time zone, asking "what is the day of this fixed_timestamp value?" doesn't even make sense -- the answer depends on where you are.

znep:
Would it make more sense to have a conversion function so you could convert fixed to floating in a specified time zone, then use existing floating timestamp date truncation functions on that?  I'm assuming the goal here is allowing server side grouping.

rjmac:
Yes, that would be ideal!  I know postgres supports that sort of thing, too (I'm almost positive it uses the system timezone database for doing the mapping, so it'd be kept up to date with changing timezone automatically), though it makes certain kinds of entirely reasonable group-by queries "interesting" from a cost perspective.  More grist for the "we need more customizable indexing" mill.

PR #36, opened 2015-12-23
Add location type for OBE.

chitang:


chrismetcalf:
This is just for OBE endpoints?

chitang:
I guess not but not sure.  But I believe we will some how only allow this to be created for internal migration use.

chrismetcalf:
@chitang OK, I'd rather not have confusion about what kind of geo columns we support in NBE/2.1

PR #42, opened 2016-01-11
Add coalesce function.

chitang:


chrismetcalf:
Got an example of how I could call this? Just like a normal SQL coalesce?

PR #45, opened 2016-02-22
Simplify typechecking

rjmac:
By removing `SoQLAnalysisType` and making typechecking choose among possibilities rather than committing to a simple type too soon.


chitang:
lgtm

PR #47, opened 2016-03-23
Add SoQL functions point.latitude, point.longitude

chitang:


chrismetcalf:
Yay! :+1: 

When should I look for this in production?

meganschneider:
:shipit:

PR #48, opened 2016-03-24
Nathaliek/en 3726

NathalieKaligirwa:
Notes: 
- geo related functions should start with geo 
- added code path to make a polygon valid and make a type a "multi" type. To be used with caution because the return type is supposed to be the same as the column the function is applied on.


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -62,6 +62,11 @@ object SoQLFunctions {
     Seq(VariableType("a"), VariableType("b")), Seq.empty, FixedType(SoQLBoolean))
   val DistanceInMeters = f("distance_in_meters", FunctionName("distance_in_meters"), Map("a" -> GeospatialLike, "b" -> GeospatialLike),
     Seq(VariableType("a"), VariableType("b")), Seq.empty, FixedType(SoQLNumber))
+  val GeoMakeValid = f("geo_make_valid", FunctionName("geo_make_valid"), Map("a" -> GeospatialLike),
+    Seq(VariableType("a")), Seq.empty, FixedType(SoQLMultiPolygon))
```
AEscobarCruz: are we still requiring that the output type be fixed here?

NathalieKaligirwa: Good point, it should be `variableType` of whatever column `a` is. Which means whoever uses `GeoMakeValid` has to perform the operation on a column with the desired output type.

AEscobarCruz:
:shipit:  once you address the one issue.

PR #50, opened 2016-03-29
Allow self-shadowing of columns

rjmac:
e.g., `select location(loc, this, that, theother) as loc`

Arbitrary circular references are still disallowed; only self-shadowing
is permitted.


chitang:
:shipit:

PR #58, opened 2016-06-29
EN-6878 Add stddev function.

chitang:


rjmac:
Ok, though aren't there supposed to be two (one for population stddev, one for sample stddev)?

chitang:
added the missing variants.

chrismetcalf:
Whoa! A `stddev` function?! Awesomesauce! 🎉

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -112,6 +112,8 @@ object SoQLFunctions {
   val Sum = f("sum", FunctionName("sum"), Map("a" -> NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"), isAggregate = true)
   val Avg = f("avg", FunctionName("avg"), Map("a" -> NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"), isAggregate = true)
   val Stddev = f("stddev", FunctionName("stddev"), Map("a" -> NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"), isAggregate = true)
```
rjmac: I don't think we need this one as a separate thing; postgres has all three but describes this one as a "historical alias for `stddev_samp`".  Since we have no history, we should probably leave it out.

rjmac:
:+1:

PR #60, opened 2016-10-30
remove cloudbees reference

JoeNunnelley:
@chitang please review and build


JoeNunnelley:
we've decided to abandon this work for now.

PR #61, opened 2017-03-24
EN-14996 Add “Distinct” support

chitang:
Just the one in the select level.

SELECT DISTINCT col1, col2, …

not the one inside function

SELECT COUNT(DISTINCT col1)

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
@@ -84,7 +84,8 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
   }
 
   private class Deserializer(in: CodedInputStream,
-                             dictionary: DeserializationDictionary[C, T])
+                             dictionary: DeserializationDictionary[C, T],
+                             version: Int)
   {
     def readPosition(): Position =
       in.readRawByte() match {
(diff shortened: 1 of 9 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
```diff
@@ -95,10 +95,15 @@ abstract class AbstractParser extends Parsers with PackratParsers {
...
-      case s ~ w ~ gb ~ h ~ ((ord, sr)) ~ ((lim, off)) => Select(s, w, gb, h, ord, lim, off, sr)
+    SELECT() ~> distinct ~ selectList ~ opt(whereClause) ~ opt(groupByClause) ~ opt(havingClause) ~ orderByAndSearch ~ limitOffset ^^ {
+      case d ~ s ~ w ~ gb ~ h ~ ((ord, sr)) ~ ((lim, off)) => Select(d, s, w, gb, h, ord, lim, off, sr)
     }
 
+  def distinct: Parser[Boolean] = opt(DISTINCT()) ^^ {
+    case Some(_) => true
+    case None => false
+  }
+
```
rjmac: `opt(DISTINCT()) ^^ (_.isDefined)` would be shorter.  I'm not insistent upon it if you like this way better though.

soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
```diff
@@ -204,6 +208,7 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
       val search = readSearch()
       SoQLAnalysis(
         isGrouped,
+        distinct,
```
rjmac: This should bump the version and add a case to `apply` down below (so that deployment is "release pg secondaries first, then query coordinator" without needing a downtime).

chrismetcalf:
Sweet! Me likey!

PR #64, opened 2017-04-27
SI-5069 Add median function.  Apps team, performance/govstats have a …

chitang:
…need for this.


chrismetcalf:
When do we expect this to get merged and are there any limitations on its availability to developers?

PR #65, opened 2017-05-03
SoQL Join

chitang:
Qualified star selection.
Use @aaaa-aaaa.field_name syntax.
support join table alias

Automatically add prefix ‘_’ in the parse tree so that end user can address resource name with just the 4x4 and no prefixes despite that the real resource name is always prefixed with ‘_’.

### DIFF
--- a/soql-environment/src/main/scala/com/socrata/soql/environment/AbstractName.scala
+++ b/soql-environment/src/main/scala/com/socrata/soql/environment/AbstractName.scala
@@ -2,8 +2,8 @@ package com.socrata.soql.environment
 
 import com.ibm.icu.text.{Normalizer2, Normalizer}
 
-abstract class AbstractName[Self <: AbstractName[Self]](val name: String) extends Ordered[Self] {
-  final lazy val caseFolded = AbstractName.caseFold(name)
+abstract class AbstractName[Self <: AbstractName[Self]](val qualifier: Option[String], val name: String) extends Ordered[Self] {
+  final lazy val caseFolded = AbstractName.caseFold(toString)
 
   protected def hashCodeSeed: Int
   override lazy val hashCode = caseFolded.hashCode ^ hashCodeSeed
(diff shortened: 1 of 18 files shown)


soql-standalone-parser/src/main/jflex/SoQLLexer.flex
```diff
@@ -126,6 +131,9 @@ QuotedSystemIdentifier = ":" "@"? ("-" | [:jletterdigit:])+
   {SystemIdentifier} { return token(new SystemIdentifier(yytext(), false)); }
   "`" / ":" { stringStart = pos(); yybegin(QUOTEDSYSTEMIDENTIFIER); }
 
+  {TableIdentifier} { return token(new TableIdentifier(yytext())); }
+  "" { stringStart = pos(); yybegin(QUOTEDIDENTIFIER); }
```
rjmac: This second line doesn't need to be here; it says that all input sequences are potentially starts of the QUOTEDIDENTIFIER state, which is definitely wrong.

soql-environment/src/main/scala/com/socrata/soql/environment/AbstractName.scala
```diff
@@ -2,8 +2,8 @@ package com.socrata.soql.environment
 
 import com.ibm.icu.text.{Normalizer2, Normalizer}
 
-abstract class AbstractName[Self <: AbstractName[Self]](val name: String) extends Ordered[Self] {
-  final lazy val caseFolded = AbstractName.caseFold(name)
+abstract class AbstractName[Self <: AbstractName[Self]](val qualifier: Option[String], val name: String) extends Ordered[Self] {
+  final lazy val caseFolded = AbstractName.caseFold(toString)
```
rjmac: I don't think ColumnName should grow a qualifier on its own; I think the qualifier should live in ColumnRef instead and keep ColumnName as just a case-insensitive identifier.

PR #66, opened 2017-05-04
SoQL Join (Qualifier is moved from ColumnName to ColumnRef)

chitang:
Basic join support
Qualified star selection
Use @aaaa-aaaa.field_name syntax
Join table alias

Automatically add prefix ‘_’ in the parse tree so that end user can address resource name with just the 4x4 and no prefixes despite that the real resource name is always prefixed with ‘_’.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
@@ -11,6 +11,7 @@ import com.socrata.soql.parsing.Parser
 import com.socrata.soql.typechecker._
 import com.socrata.soql.environment.{ColumnName, DatasetContext, TableName, TypeName}
 import com.socrata.soql.collection.OrderedMap
+import com.socrata.soql.typed.Qualifier
 
 class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Type]) {
   type Analysis = SoQLAnalysis[ColumnName, Type]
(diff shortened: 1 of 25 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -100,15 +119,18 @@ class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Ty
...
       else sourceFrom.last
 
-    def dispatch(distinct: Boolean, selection: Option[Selection], where: Option[Expression], groupBy: Option[Seq[Expression]], having: Option[Expression], orderBy: Option[Seq[OrderBy]], limit: Option[BigInt], offset: Option[BigInt], search: Option[String]) =
+    def dispatch(distinct: Boolean, selection: Option[Selection],
+                 joins: Option[List[Tuple2[TableName, Expression]]],
+                 where: Option[Expression], groupBy: Option[Seq[Expression]], having: Option[Expression], orderBy: Option[Seq[OrderBy]], limit: Option[BigInt], offset: Option[BigInt], search: Option[String]) =
       selection match {
-        case None => analyzeNoSelectionInOuterSelectionContext(lastQuery, distinct, where, groupBy, having, orderBy, limit, offset, search)
-        case Some(s) => analyzeInOuterSelectionContext(lastQuery, Select(distinct, s, where, groupBy, having, orderBy, limit, offset, search))
+        case None => analyzeNoSelectionInOuterSelectionContext(lastQuery, distinct, None, where, groupBy, having, orderBy, limit, offset, search)
```
rjmac: Not `joins`?

soql-standalone-parser/src/main/jflex/SoQLLexer.flex
```diff
@@ -126,6 +131,8 @@ QuotedSystemIdentifier = ":" "@"? ("-" | [:jletterdigit:])+
   {SystemIdentifier} { return token(new SystemIdentifier(yytext(), false)); }
   "`" / ":" { stringStart = pos(); yybegin(QUOTEDSYSTEMIDENTIFIER); }
 
+  {TableIdentifier} { return token(new TableIdentifier(yytext())); }
```
rjmac: If you don't want the `@` to be a part of the name (I see it gets stripped off in the parser), you can `.substring(1)` the yytext here.

chitang: Good to know.  But I'll leave it in AbstractParser.

PR #67, opened 2017-05-10
EN-16084 Add resource name

chitang:


chrismetcalf:
@chitang @rjmac Que es eso? Does this have any affects that'd be interesting to me?

PR #72, opened 2017-05-24
Make joins optional at the parse level

rjmac:
DSMAPI doesn't need these and we can get a better error message if
trying to use them is a parse error.


chitang:
:shipit:

PR #73, opened 2017-06-13
EN-16372 Supports outer joins.

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/typed/Join.scala
```diff
@@ -0,0 +1,39 @@
...
+object Join {
+
+  def apply[ColumnId, Type](joinType: String, tableName: TableName, expr: CoreExpr[ColumnId, Type]): Join[ColumnId, Type] = {
+    joinType match {
+      case InnerJoinName => InnerJoin(tableName, expr)
+      case LeftOuterJoinName => LeftOuterJoin(tableName, expr)
+      case RightOuterJoinName => RightOuterJoin(tableName, expr)
+      case jt: String => throw new IllegalArgumentException(s"invalid join type $jt")
+    }
+  }
```
rjmac: Not a fan of this function.  I don't think join types should be arbitrary strings.

PR #74, opened 2017-06-15
EN-16454 Support multiple “SELECT *”’s

chitang:
There can be * from more than one table.  But :* is still limited to one table because system columns always contain name collisions.


chitang:
Yeah, we were fighting the "_" prefix problem from core when we started joins.  Cleaner solutions would require massive backfill or much more work if we do not apply these tricks low enough.

PR #77, opened 2017-06-26
Chi/en 16370/sub queries

chitang:
Partial sub-queries support to allow processing joins with derived views.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisSerializer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisSerializer.scala
@@ -220,8 +220,8 @@ class AnalysisSerializer[C,T](serializeColumn: C => String, serializeType: T => 
       out.writeUInt32NoTag(size)
       joins.toList.flatten.foreach { join =>
         out.writeStringNoTag(join.typ.toString)
-        out.writeStringNoTag(join.tableName.name)
-        maybeWrite(join.tableName.alias)(alias => out.writeStringNoTag(alias))
+        write(join.tableLike)
+        maybeWrite(join.alias)(x => out.writeStringNoTag(x))
         writeExpr(join.expr)
       }
     }
@@ -278,6 +278,7 @@ class AnalysisSerializer[C,T](serializeColumn: C => String, serializeType: T => 
       val SoQLAnalysis(isGrouped,
                        distinct,
                        selection,
+                       from,
                        join,
                        where,
                        groupBy,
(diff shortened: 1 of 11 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/AnalysisSerializer.scala
```diff
@@ -289,6 +290,7 @@ class AnalysisSerializer[C,T](serializeColumn: C => String, serializeType: T =>
       writeGrouped(isGrouped)
       writeDistinct(analysis.distinct)
       writeSelection(selection)
+      maybeWrite(from) { x => out.writeStringNoTag(x) }
```
rjmac: You'll want to do the version-bumping dance here and in the deserializer

rjmac:
It is not immediately obvious to me what will happen if labels are reused inside subqueries, e.g., "... join (select a_col from aaaa-aaaa join bbbb-bbbb as x where a_col=x.col) as x where my_col = x.a_col"

soql-analyzer/src/main/scala/com/socrata/soql/AnalysisSerializer.scala
chitang: Joins aren't really in active use.  Therefore we can avoid the dance.  We all ignored version difference when we added outer joins.

chitang:
Do you mean serialization dictionary collision when you say labels reused inside subqueries?

chitang:
You mean both aaaa-aaaa and bbbb-bbbb have the same column name?

rjmac:
I mean the inner and outer queries both alias something to `x`

chitang:
i will look.  But I think we don't need to handle nested joins yet.

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -416,6 +418,57 @@ case class SoQLAnalysis[ColumnId, Type](isGrouped: Boolean,
...
+      orderBy = orderBy.map(_.map(_.mapColumnIds(t2ToF2(qColumnIdNewColumnIdWithJoinsMap))))
+    )
+  }
+
+  /**
+    * Change function from taking tuple2 to function taking two parameters.
+    */
+  private def t2ToF2[A, B, X](f: Tuple2[A, B] => X)(a: A, b: B): X = {
+    f(Tuple2(a, b))
+  }
```
rjmac: This is `Function.untupled`

chitang:
I'll remove/replace it next time.  Thanks

chitang:
I might have mis-read an email and mistakenly thought that you approved the new commit.

PR #79, opened 2017-07-12
EN-17515 support RowId being Long vice BigDecimal.  Otherwise, query is not using the index.

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
```diff
@@ -137,6 +137,10 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
             readExpr()
           }
           FunctionCall(func, params)(pos, functionNamePosition)
+        case 7 =>
+          val value = java.lang.Long.parseLong(dictionary.strings(in.readUInt32()))
+          val typ = dictionary.types(in.readUInt32())
```
rjmac: Longs are 64-bit, not 32

rjmac: oh wait nm, I see what you're doing.  but: you should probably just write the int instead of converting it to a string and saving a reference to that.

chitang:
Yeah, I think I can just do the trick in soql-pg without touching soql-reference.

PR #86, opened 2017-09-27
Update type checking to accommodate window functions

chitang:


soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
```diff
@@ -501,4 +501,21 @@ SELECT visits, @x2.zx
       ColumnName("zx") -> ColumnRef(Some("_x2"), ColumnName("zx"), TestText)(NoPosition)
     ))
   }
+
+  test("window function") {
+    val analysisWordStyle = analyzer.analyzeUnchainedQuery("SELECT avg(visits) OVER (PARTITION BY name_last, 2, 3)")
+    val select = analysisWordStyle.selection.toSeq
+    select must equal (Seq(
+      ColumnName("avg_visits_name_last_2_3") -> typedExpression("avg(visits, name_last, 2, 3)")
```
rjmac: This is _really_ weird to me.  Can you explain why you chose this analysis result?

rjmac: In particular, why not have this compile into something like `op$over(avg(visits), name_last, 2, 3)` and then have the `op$over` "function" be a form treated specially by the sql interpreter?  As it is, it looks like it'd be easy for users to accidentally create a window call just by incorrectly passing multiple parameters to `avg`.

chitang: I am not attached to the specific result.  op$over does look less weird.

soql-analyzer/src/main/scala/com/socrata/soql/functions/MonomorphicFunction.scala
```diff
@@ -6,7 +6,7 @@ case class MonomorphicFunction[Type](function: Function[Type], bindings: Map[Str
   def this(identity: String, name: FunctionName, parameters: Seq[Type], repeated: Seq[Type], result: Type, isAggregate: Boolean = false) =
     this(Function(identity, name, Map.empty, parameters.map(FixedType(_)), repeated.map(FixedType(_)), FixedType(result), isAggregate), Map.empty)
 
-  require(bindings.keySet == function.parameters.collect { case VariableType(n) => n }.toSet, "bindings do not match")
+  require(bindings.keySet.filterNot(_ == "?") == function.parameters.collect { case VariableType(n) => n }.toSet, "bindings do not match")
```
rjmac: It'd probably be better to create a new subclass of `TypeLike` rather than treating `VariableType("?")` specially

chitang: Thought about that too.  Yeah, that is better.

soql-analyzer/src/test/scala/com/socrata/soql/types/TestFunctions.scala
```diff
@@ -34,14 +34,16 @@ object TestFunctions {
...
-              Map("a" -> Set(TestNumber), "?" -> AllTypes),
-              Seq(VariableType("a")),
-              Seq(VariableType("?")),
-              VariableType("a"),
-              isAggregate = true)
+  val Avg = f("avg", FunctionName("avg"), Map("a" -> NumLike), Seq(VariableType("a")), Seq.empty, VariableType("a"), isAggregate = true)
   val CountStar = mf("count(*)", SpecialFunctions.StarFunc("count"), Seq(), Seq.empty, TestNumber, isAggregate = true)
 
+  val WindowFunctionOver = f("wf_over",
+                             FunctionName("wf_over"),
```
rjmac: A name from `SpecialFunctions` would seem appropriate here, if you don't want people actually writing `wf_over(...)`

rjmac: (that'd also provide a reasonable hook for printing it back as window-function syntax)

soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
```diff
@@ -506,13 +506,13 @@ SELECT visits, @x2.zx
...
     select must equal (Seq(
-      ColumnName("wf_over_avg_visits_name_last_2_3") -> typedExpression("wf_over(avg(visits), name_last, 2, 3)")
+      ColumnName("wf_over_avg_visits_name_last_2_3") -> typedExpression("avg(visits) OVER (PARTITION BY name_last, 2, 3)")
     ))
 
     val analysisWordStyleOverEmpty = analyzer.analyzeUnchainedQuery("SELECT avg(visits) OVER ()")
     val selectOverEmpty = analysisWordStyleOverEmpty.selection.toSeq
     selectOverEmpty must equal (Seq(
-      ColumnName("wf_over_avg_visits") -> typedExpression("wf_over(avg(visits))")
+      ColumnName("wf_over_avg_visits") -> typedExpression("avg(visits) OVER ()")
```
rjmac: I just approved it, but I'm not sure this column name is the right thing to generate.  Can you see if it'll be plausible to generate avg_visits_over or the like?

PR #92, opened 2017-11-14
EN-19155: add geo collection extract functions

bahrieinn:
Some `ST_MakeValid` calls return `GEOMETRYCOLLECTION` which we don't handle, so we need to call `ST_CollectionExtract` to turn it back into `MULTIPOLYGON`s


chitang:
I wonder if you just want one function with variable binding

PR #97, opened 2018-01-13
Prevent a combinatorial explosion in the typechecker

rjmac:


soql-analyzer/src/main/scala/com/socrata/soql/typechecker/Typechecker.scala
```diff
@@ -78,7 +78,17 @@ class Typechecker[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Typ
...
           }.map(typed.FunctionCall(f, _)(fc.position, fc.functionNamePosition))
-        })
+        }
+
+        // If all possibilities result in the same type, we can disambiguate here.
+        // In principle, this could be
+        //   potentials.groupBy(_.typ).values.map(disambiguate)
+        // which would be strictly more general, but I am uncertain whether
+        // that will preserve the preference-order assumption that disambiguate
+        // relies on.
```
rjmac: (in fact I know it will not)

chitang:
I add a test to capture the problem in the branch chi/add-test.  How useful is it to have concat be polymorphic?

rjmac:
Pretty useful, at least hypothetically, in the same way any2stringadd is useful in Scala.  In any case, with this change all implicit conversions will be resolved on the way up the tree, so it's not really an issue unless we have something that's polymorphic in at least one argument and polymorphic using a _different_ variable in its result type.  In that case, it'd be worth doing the work to make the `groupBy`/`map` line I referenced in my comment work (it wouldn't be hard, it'd just be an extra `sort` step between `values` and `map`).

Your branch looks good, if you could merge it into this branch so it all goes in as one PR that'd be great.

PR #100, opened 2018-03-29
EN-23592 system column names in column aliases

chitang:
So that we can do max(:id) as :id.  This help interoperability with obe

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
@@ -9,7 +9,7 @@ import com.socrata.soql.ast._
 import com.socrata.soql.environment.{ColumnName, FunctionName, TableName, TypeName}
 
 object AbstractParser {
-  class Parameters(val allowJoins: Boolean = true, val systemColumnAliasesAllowed: Set[String] = Set.empty)
+  class Parameters(val allowJoins: Boolean = true, val systemColumnAliasesAllowed: Set[ColumnName] = Set.empty)
   val defaultParameters = new Parameters()
 }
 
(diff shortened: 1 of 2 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
```diff
@@ -200,9 +200,16 @@ abstract class AbstractParser(parameters: AbstractParser.Parameters = AbstractPa
       }
     }
 
-  def namedSelection = expr ~ opt(AS() ~> simpleUserIdentifier) ^^ {
+  private val standardSystemColumns = Set(":id", ":created_at", ":updated_at")
```
rjmac: Not a fan of this; if you're going to do it this way, have the set of standard system columns passed in as a parameter when the parser is constructed.

chitang: Yeah, that will be cleaner.  I will update it.

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
```diff
@@ -9,7 +9,7 @@ import com.socrata.soql.ast._
 import com.socrata.soql.environment.{ColumnName, FunctionName, TableName, TypeName}
 
 object AbstractParser {
-  class Parameters(val allowJoins: Boolean = true)
+  class Parameters(val allowJoins: Boolean = true, val systemColumnAliasesAllowed: Set[String] = Set.empty)
```
rjmac: Can this be a `Set[ColumnName]` instead?  Only because user code will probably have them in `ColumnName` form already, and moving the `ColumnName(name)` up two lines down below in `namedSelection` is simpler than `systemColumns.map(_.underlying)`

PR #107, opened 2018-09-21
ast update

EddieCarlsonSocrata:
The changes here aim to simplify the definitions of the SoQL AST in the parser and analyzer, as well as having the data structures more closely, clearly, and accurately represent the things we want to do with SoQL, while disallowing the representation of invalid soql.

Some of the changes include:
1) removing "from" from Select - the only from that can show up is in a join, since a top-level select is implicitly "from" the current view.
2) improve the join class to make it less confusing to determine the source of the join as well as the alias and potential subqueries in the join. 
3) change joins, groupBys, and orderBys from Option[Seq[Type]] to Seq[Type] - more correct (Some(Nil) AND None are both isomorphic to Nil) and quality of life improvement (don't have to flatten/nest maps all the time)
4) improve aliasing
5) functional-style improvements
6) supporting changes in the parser
7) mirrored changes in analyzer
8) added comments
9) use NonEmptySeq when at least one element is required (top-level parsing and join sub-selects/analyses)

Potential issues:
1) Do we need to use the Analysis deserializer to deserialize any Analysis objects serialized using the previous version of the serializer? if so, need to add back in a path for that.

Potential future improvements:
1) Adding ability for users to write nested subqueries in joins. joins support chained soql views as a subquery, but the user can't write multiple froms in the soql editor. A user can write "join (select a, b from @aaaa-aaaa) as a", but not "join (select a, b from (select a, b, c from @aaaa-aaaa) as a1) as a". This is not a regression from previous behavior, was disallowed before as well. This will require reconciling the current way of representing chained soql (which implies a "from" relationship) with a user-written soql nested from (which may have additional aliasing information). It might be confusing/invalid to have multiple representations (a chained soql element of index other than 0 also containing a "from" representing another subquery). 
2) Possibly reconciling Analysis with Select objects. Will be difficult because of the differences in typing, particularly around CoreExpr/Expression. Possible to do, but when I explored various means of doing this, it began to feel very cumbersome.

Other Notes:
Rolled my own NonEmptySeq class instead of using scalaz's NonEmptyList for two reasons: 1) scalaz is a big dependency to pull in just for the one class 2) a socrata cultural/stylistic preference for Seq over List.


rjmac:
Not a fan of all the `Seq` to `List` conversions.  Every time I see a `List` I expect that somewhere something is going to be structural recursion on cons cells, and that never happens here.  Sometimes the head is treated specially, but that's it.

rjmac:
(also the "sometimes the head is treated specially" makes me wonder if you want `List[T]` at all and not `Something(T1, others: Seq[T2])` but I haven't thought super hard about that.  It just jumps out at me as a repeated pattern)

rjmac:
Having looked a little more closely, yeah I think that's what you want -- if scalac were a little smarter it would have given you partial pattern match warnings on most of the places where you're pulling off the head like that, as you're not matching the empty case.

chitang:
Does dis-allowing top level from interfere with future possible "union" feature like
select a union select b from t_b?

chitang:
I think the convention is to use () on functions with side effect.

EddieCarlsonSocrata:
@rjmac I did notice one potential pattern match error in typed.Join, fixed it in a commit just now. I did consider using NonEmptyList for the SubSelect/SubAnalysis class (and maybe also for top level returns from Analyzer/Parser). I can see the argument for using NonEmptyList - not sure what you're getting at with the "Something" class.

EddieCarlsonSocrata:
@chitang I don't think this disallows union - certainly not any more than it used to anyway. Union might be implemented as a class that wraps two Selects/Analyses. Also, I've never followed the "()" implies side effect convention, but if that's something the team wants to do, i'm happy to go add the ()s back in.

chitang:
Agree with the union part.  Please put () back to functions with side effect.

EddieCarlsonSocrata:
@chitang done!

EddieCarlsonSocrata:
@rjmac how do you feel about the implementation of NonEmptySeq and its usage to replace top-level parses and Join sub-selects/analyses? (soql-standalone-parser/src/main/scala/com/socrata/NonEmptySeq.scala)

rjmac:
@EddieCarlsonSocrata Definitely :+1: on more precise types.  It looks really good to me.

PR #108, opened 2018-10-15
EN-27838: don't explode on strings that look like very large numbers

rjmac:


chitang:
:shipit: what actual exception will that trigger?

rjmac:
NumberFormatException.  But actually, since the right fix was so straightforward I'll release a new RJ tonight and just bump the required version.

PR #111, opened 2018-11-01
more robust prefixing support (for both 4x4s and aliases) in TableName

EddieCarlsonSocrata:


chitang:
We have intentionally not got into core 4x4 in all nbe services.

PR #113, opened 2018-11-30
En 27882/joined column alias

EddieCarlsonSocrata:
default computed column names on joined columns should contain joined table name/alias

### DIFF
--- a/soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
+++ b/soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
@@ -560,4 +560,10 @@ SELECT visits, @x2.zx
         (NoPosition, NoPosition)
     ))
   }
+
+  test("alias is part of computed function name") {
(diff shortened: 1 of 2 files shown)


soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
```diff
@@ -560,4 +560,10 @@ SELECT visits, @x2.zx
         (NoPosition, NoPosition)
     ))
   }
+
+  test("joined table name alias is part of computed function name") {
```
chitang: I think it is better to remove the word "computed".  We use that word mostly for computed columns.

chitang:
The rest lgtm, :shipit: if you make the change.

soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
chitang: maybe part of expression name

chitang:
:shipit:

PR #117, opened 2019-04-08
EN-31669 Window function support, phase 2

chitang:
1. Support order by
2. Add row_number, rank, dense_rank

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
@@ -46,8 +46,7 @@ object Expression {
         case FunctionCall(SpecialFunctions.WindowFunctionOver, args) =>
           Seq(args.head).flatMap(findIdentsAndLiterals) ++
             Vector("over") ++
-            (if (args.tail.isEmpty) Vector.empty else Vector("partition", "by")) ++
-            args.tail.flatMap(findIdentsAndLiterals)
+            (if (args.tail.isEmpty) Vector.empty else args.tail.flatMap(findIdentsAndLiterals))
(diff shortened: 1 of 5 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -46,8 +46,7 @@ object Expression {
         case FunctionCall(SpecialFunctions.WindowFunctionOver, args) =>
           Seq(args.head).flatMap(findIdentsAndLiterals) ++
```
EddieCarlsonSocrata: Actually, this whole thing should be rewritten to `val mapped = args.flatMap(findIdentsAndLiterals)
mapped.head :: "over" :: mapped.tail`

EddieCarlsonSocrata: wait, sorry, this is wrong. it should be:
`findIdentsAndLiterals(args.head) ++ "over" ++ args.tail.flatMap(findIdentsAndLiterals)`

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -199,9 +197,10 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression])
       case SpecialFunctions.NotLike =>
         parameters.map(_.format(d)).mkString(" NOT LIKE ")
       case SpecialFunctions.WindowFunctionOver =>
-        val param0 = parameters(0).format(d)
-        val partitions = windowOverPartition(parameters.tail)
-        val orders = windowOverOrder(parameters.tail)
+        val head :: tail = parameters
```
EddieCarlsonSocrata: ah, cool, this match is nicer than the one i suggested.

PR #118, opened 2019-04-10
EN-31669 Do not merge chained soul on window function

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -573,70 +573,84 @@ private class Merger[T](andFunction: MonomorphicFunction[T]) {
...
-                        groupBys = aGroup,
-                        having = mergeWhereLike(aSelect, aHaving, bWhere),
-                        orderBys = mergeOrderBy(aSelect, aOrder, bOrder),
-                        limit = bLim,
-                        offset = bOff,
-                        search = aSearch))
-    case (_, _) =>
-      None
+  private def tryMerge(a: Analysis, b: Analysis): Option[Analysis] = {
+    if (hasWindowFunction(a)) { None }
```
PiSoup: Would it make more sense to pull this into the existing match block rather than adding a second branching point? I'm thinking like just adding a basic `case (a, _) if hasWindowFunction(a) => None` to the top

PR #119, opened 2019-04-18
Fix window function parsing

rjmac:
Window function expressions are always of the form `function(args)
over (more args)` but the grammar would accept `identifier over (more
args)`

Also, the `windowFunctionParams` term could never produce a Left, so
de-either it.  This gets rid of a partial function call in the
function call term.

Also silence a warning.


chitang:
:shipit:

PR #123, opened 2019-05-29
EN-32659 Add text to photo

chitang:


chitang:
On hold, I am going to take a different path

soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala
```diff
@@ -66,6 +67,12 @@ object SoQLPackDecoder {
     case _: Exception => None
   }
 
+  def decodePhoto(item: Any): Option[SoQLValue] = try {
```
PiSoup: Could we just have `photo` use the same type as `blob` at this level of the stack, or are there other differences between them?

chitang: My first attempt tried to mix/combine the two.  I don't really like mixing them.

PR #124, opened 2019-07-04
SoQLUrl encode/decode to soqlpack

catstavi:


chitang:
While you are here, can you add SoQLDocument too?

catstavi:
@chitang unfortunately I already wrapped this up and also don't really know anything about what SoQLDocument looks like! I'll keep it in mind as something to try picking up when I'm finished with these customer escalations!

soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala
```diff
@@ -80,6 +81,18 @@ object SoQLPackDecoder {
...
 
+  def decodeUrl(item: Any): Option[SoQLValue] = item match {
+    case Seq(url: Array[Byte], label: Array[Byte]) =>
+      Some(SoQLUrl(Some(new String(url, StandardCharsets.UTF_8)), Some(new String(label, StandardCharsets.UTF_8))))
+    case Seq(url: Array[Byte], null) =>
+      Some(SoQLUrl(Some(new String(url, StandardCharsets.UTF_8)), None))
+    case Seq(null, label: Array[Byte]) =>
+      Some(SoQLUrl(None, Some(new String(label, StandardCharsets.UTF_8))))
+    case Seq(null, null) =>
+      Some(SoQLUrl(None, None))
```
rjmac: Annoyingly, this is probably the most concise way to write this.

soql-pack/src/main/scala/com.socrata.soql/SoQLPackEncoder.scala
```diff
@@ -49,6 +50,18 @@ object SoQLPackEncoder {
...
+  def encodeUrl(soqlUrl: SoQLUrl): Any = soqlUrl match {
+    case SoQLUrl(Some(url), Some(label)) =>
+      Array(url.getBytes(StandardCharsets.UTF_8), label.getBytes(StandardCharsets.UTF_8))
+    case SoQLUrl(Some(url), None) =>
+      Array(url.getBytes(StandardCharsets.UTF_8), null)
+    case SoQLUrl(None, Some(label)) =>
+      Array(null, label.getBytes(StandardCharsets.UTF_8))
+    case SoQLUrl(None, None) =>
+      Array(null, null)
+  }
```
rjmac:
```scala
def encodeUrl(soqlUrl: SoQLUrl) = {
  val SoQLUrl(url, label) = soqlUrl
  Array(url.map(_.getBytes(StandardCharsets.UTF_8)).orNull,
        label.map(_.getBytes(StandardCharsets.UTF_8)).orNull))
}

Comment (catstavi) [reply soql-pack/src/main/scala/com.socrata.soql/SoQLPackEncoder.scala]: ✨ 😮

File: soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala (hunk 1 above)
Comment (branweb1) [reply soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala]: wait why not return a None on ln 92?

Comment (branweb1) [reply soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala]: or at least have a catch-all

File: soql-pack/src/main/scala/com.socrata.soql/SoQLPackDecoder.scala [hunk 3]
```diff
@@ -79,6 +81,21 @@ object SoQLPackDecoder {
...
+        Some(SoQLUrl(Some(new String(url, StandardCharsets.UTF_8)), Some(new String(label, StandardCharsets.UTF_8))))
+      case Seq(url: Array[Byte], null) =>
+        Some(SoQLUrl(Some(new String(url, StandardCharsets.UTF_8)), None))
+      case Seq(null, label: Array[Byte]) =>
+        Some(SoQLUrl(None, Some(new String(label, StandardCharsets.UTF_8))))
+      case Seq(null, null) =>
+        Some(SoQLUrl(None, None))
+    }
+  } catch {
+    case _: Exception => None
```
branweb1: i don't think this needs to be in a try-catch. The other one was because it's trying to decode to json. Also did you figure out the ues case for a Some(SoQUrl(none, none)) over just none?

branweb1: what I mean is: it seems like it would be better to just add a case statement that matches everything to make your pattern-match exhaustive. And ideally figure out if Some(SoQLUrl(None, None)) is needed.

catstavi: I walked through this a little more, to see that the decoder producing None makes a SoQLNull for the library user, which seems totally reasonable. 👍 so I went ahead with what you're suggesting!

PR #132, opened 2019-11-12
EN-35828 Support autocast from numeric literal to text literal.

chitang:
text_column > 123 

should work like

text_column > 123::text


rjmac:
I don't like this, for a few reasons:
* There is no canonical conversion from a numeric value to a string.  E.g., there's nothing to say that the numeric literal `123` doesn't become the string `"123.000000"` or `"1.23e2"` as those are both valid representations of that value.
* In the presence of schema changes, this can leak information.  Say you have a derived view with a saved query like `where some_col < 200`.  Change `some_col` to text and suddenly this derived view is returning rows where `some_col` contains `"1000000"` instead of getting a "your view is broken" error.  *this is possibly extremely important for SCGC federation*

rjmac:
(more generally to those points: I'm very strongly of the opinion literal conversion should only be used with source type text, the target type has no literal representation of its own, and there exists only one reasonable representation of the target type within text.  E.g., ISO8601 for dates or WKT for geometries.  Numbers and booleans should _not_ have automatic conversions from text to themselves, nor should they have implicit conversions from themselves to any other types.  Anything else is just begging for confusion and potential information leakage.)

PR #137, opened 2020-01-17
EN-37621, EN-37687 date-x - date-y -> z seconds, abs(x) => y

chitang:
Specify SoQL functions:

Datediff in seconds
Absolute

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -11,7 +11,7 @@ sealed abstract class SoQLFunctions
 object SoQLFunctions {
   private val log = org.slf4j.LoggerFactory.getLogger(classOf[SoQLFunctions])
 
-  import SoQLTypeClasses.{Ordered, Equatable, NumLike, RealNumLike, GeospatialLike}
+  import SoQLTypeClasses.{Ordered, Equatable, NumLike, RealNumLike, GeospatialLike, TimestampLike}
   private val AllTypes = SoQLType.typesByName.values.toSet
 
   // helpers to guide type inference (specifically forces SoQLType to be inferred)
@@ -145,7 +145,6 @@ object SoQLFunctions {
 
   val BinaryPlus = f("+", SpecialFunctions.Operator("+"), Map("a" -> NumLike), Seq(VariableType("a"), VariableType("a")), Seq.empty, VariableType("a"))
   val BinaryMinus = f("-", SpecialFunctions.Operator("-"), Map("a" -> NumLike), Seq(VariableType("a"), VariableType("a")), Seq.empty, VariableType("a"))
-  val FloatingTimestampMinus = mf("floating timestamp -", SpecialFunctions.Operator("-"), Seq(SoQLFloatingTimestamp, SoQLFloatingTimestamp), Seq.empty, SoQLNumber)
 
   val TimesNumNum = mf("*NN", SpecialFunctions.Operator("*"), Seq(SoQLNumber, SoQLNumber), Seq.empty, SoQLNumber)
   val TimesDoubleDouble = mf("*DD", SpecialFunctions.Operator("*"), Seq(SoQLDouble, SoQLDouble), Seq.empty, SoQLDouble)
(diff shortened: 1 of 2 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -145,6 +145,7 @@ object SoQLFunctions {
 
   val BinaryPlus = f("+", SpecialFunctions.Operator("+"), Map("a" -> NumLike), Seq(VariableType("a"), VariableType("a")), Seq.empty, VariableType("a"))
   val BinaryMinus = f("-", SpecialFunctions.Operator("-"), Map("a" -> NumLike), Seq(VariableType("a"), VariableType("a")), Seq.empty, VariableType("a"))
+  val FloatingTimestampMinus = mf("floating timestamp -", SpecialFunctions.Operator("-"), Seq(SoQLFloatingTimestamp, SoQLFloatingTimestamp), Seq.empty, SoQLNumber)
```
rjmac: This will give nonsense answers a significant fraction of the time.

rjmac: (because it doesn't and cannot take into account DST shifts.  Put this operator on fixed timestamp and it will be fine.)

chitang: We have very few fixed timestamp columns.  Doesn't it still make sense w/o time zone?  It can be wrong if people assume floating timestamp is fixed timestamp in a specific time zone.  But even that the error is contained and understood?  I think we may want to add two more functions though 1. the same function in fixed timestamp.  2. a function that takes a timezone and convert a floating timestamp to a fixed timestamp.

chitang: I don't see better way to help people do this on columns that are already in floating_timestamp (the majority).

PiSoup:
If we're adding timestamp math is there value in also adding an 'interval' type?

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
rjmac: It does not make sense without a time zone.  How many seconds are between noon on march 7th and noon on march 9th, in 2020?  The answer is: it depends on which time zone you're representing those times in.  If they're with respect to Seattle time, it's 47 hours, 169200 seconds.  If they're with respect to Tucson time, it's 48 hours, 172800 seconds.

Unfortunately, it's not actually possible to do this calculation in general given only a floating timestamp and a time zone, because the conversion between an instant and a wall-clock representation is lossy.  How many seconds are there between midnight at the start of the 1st of november 2020 and 1:30am on that same day, in Seattle specifically (so with the time zone "US/Pacific")?  The answer is: either 5400 seconds or 9000 seconds depending on whether you mean the first 1:30am on that day or the second 1:30am on that day, since it happens twice, and `floating_timestamp` does not represent that.

Really almost all times on the platform ought to be `fixed_timestamp`, with `floating_timestamp` reserved for presentation purposes ("I as a dataset owner want the times in this dataset to be displayed in New York time, not the user's local timezone, so I shall create a view with either `select ... to_floating_timestamp(my_time_column, 'US/Eastern') ...`" or a formatting option on the column that makes the frontend do that transformation either internally to itself or as a transformation on the soql it emits when populating its internal data.

PiSoup: Perhaps if we only add math functions to fixed_timestamp and improve column formatting information so that it can include timezone or whatever we could entice people to use fixed_timestamp. Or make the migration ourselves.

rjmac:
There absolutely is!  Indeed there's value in adding _two_ (corresponding to the notions usually called "duration" and "period" -- "duration" is a fixed amount of time e.g., some number of seconds, and "period" is a change with respect to some calendar (e.g., "1 day ahead", which can be 23, 24, or 25 hours but is always a bump to the "days" column of the type we call "floating timestamp").

rjmac:
and, since our customers are frequently interested in only the "date" part of timestamps, there's value in adding separate "date" and "time" types too -- fun fact, these are both actually already implemented in the NBE (or at least in truth), but are not exposed and have no soql library support that I'm aware of.

rjmac:
(also also there'd be value in adding types that correspond to Java's `ZonedDateTime` and `OffsetDateTime` types, but I'm not actually sure how those would map onto postgresql)

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLTypeClasses.scala
```diff
@@ -30,4 +30,5 @@ object SoQLTypeClasses {
 
   val NumLike = Set[SoQLType](SoQLNumber, SoQLDouble, SoQLMoney)
   val RealNumLike = Set[SoQLType](SoQLNumber, SoQLDouble)
+  val TimestampLike = Set[SoQLType](SoQLFloatingTimestamp, SoQLFloatingTimestamp)
```
rjmac: I think there's some mistake here...

PR #138, opened 2020-02-05
EN-38087 support SELECT COUNT DISTINCT column

chitang:


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -189,6 +190,8 @@ object SoQLFunctions {
   val RowNumber = mf("row_number", FunctionName("row_number"), Seq(), Seq.empty, SoQLNumber)
   val Rank = mf("rank", FunctionName("rank"), Seq(), Seq.empty, SoQLNumber)
   val DenseRank = mf("dense_rank", FunctionName("dense_rank"), Seq(), Seq.empty, SoQLNumber)
+  val FirstValue = f("first value", FunctionName("first_value"), Map.empty, Seq(VariableType("a")), Seq.empty, VariableType("a"))
```
vickingan: These don't seem to be used anywhere in this PR?

chitang: They are just new functions (declaration).  They will be used in soql-postgres-adapter

PR #142, opened 2020-04-14
Include full text search string when rendering an AST back into a string

rozap:


znep:
Hrm, I wonder if making it multi line will break anything ... is that just a style change or related to whatever the actual issue is?

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -87,17 +87,17 @@ case class Select(
...
-      val parts = List(selectStr, fromStr, joinsStr, whereStr, groupByStr, havingStr, obStr, limitStr, offsetStr, searchStr)
+      val fromStr = from.map(t => s"\nFROM $t")
+      val joinsStr = itrToString(None, joins.map(j => s"\n${j.toString}"), "\n ")
+      val whereStr = itrToString("\nWHERE", where)
+      val groupByStr = itrToString("\nGROUP BY", groupBys, ", ")
+      val havingStr = itrToString("\nHAVING", having)
+      val obStr = itrToString("\nORDER BY", orderBys, ", ")
+      val limitStr = itrToString("\nLIMIT", limit)
+      val offsetStr = itrToString("\nOFFSET", offset)
+      val searchStr = itrToString("\nSEARCH", search.map(Expression.escapeString))
```
chitang, line 98: Can we move searchStr up before limitStr so that it follows the same logical order?

rozap:
multiline-ness is a style change, unrelated to this actual issue. but it's related to the `/explore` page in that, when you fiddle with the query through some UI element like a column sort dropdown, it updates the text of the query, and it'd be preferable if it didn't mush it into one single line thing on each interaction. 

that problem in general would be interesting to think more about, but i haven't done it, so the next best thing was to make the default pretty print at least sort of readable for that case.

znep:
What I'm saying is other things that rely on turning it into a string may or may not get sad if suddenly the output is multiline.  I know rollup code depends on this, no idea if making it multiline will break anything, not sure what else does.

znep:
I'm also not sure on what triggers the pretty path vs not-pretty offhand or what the difference is.

rjmac:
Nothing; the boolean is a constant.  The difference is that the not-pretty path shows the raw case classes in their ordinary auto-generated format.

This was a mistake, I should have just made them separate methods.

rozap:
i wonder if, in that case, it'd be worth creating a `format()` function that returns the pretty string subject to whatever formatting changes we make as part of this project going forward? And then it's explicit rather than part of `toString()` ? And then i'll revert the toString implementation back to how it was?

PR #144, opened 2020-06-24
EN-40709: Support frame clause in window functions

chitang:


soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -345,6 +353,8 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression])
     }
     val ps1 = ps.takeWhile {
       case StringLiteral("order_by") => false
+      case StringLiteral("range") => false
+      case StringLiteral("rows") => false
```
rjmac, line 357: I hate everything about the way this is done.  Window function calls should have their own type in the AST distinct from ordinary function calls.

rjmac: Walking down this path will just lead to a repeat of the join qualification mess, where the parser and analyzer know an important thing, but then they erase it, leaving downstream users of their output to reconstruct things that they should just be given.

chitang: address this by specialized WindowFunctionCall

chitang: Still need to work on the typed version.

PR #146, opened 2020-07-06
EN-40991: Window function refactor part II (part I EN-40709)

chitang:
. Use the new style in analysis serializer
. Remove old serialization stuff
. Refine window function syntax

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
@@ -227,8 +227,7 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
         dictionary.strings(in.readUInt32())
       }
 
-    def readAnaAndFrom(): SoQLAnalysis[C, T] = {
-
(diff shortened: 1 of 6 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
```diff
@@ -266,61 +227,43 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
...
 
-      val ana = SoQLAnalysis(ig, d, s, j, w, gb, h, ob, l, o, search)
-
-      (ana, fromOptTrigger)
+      SoQLAnalysis(ig, d, s, j, w, gb, h, ob, l, o, search)
     }
 
     def readAnalysis(): SoQLAnalysis[C, T] = {
-      readAnaAndFrom()._1
+      readAnaAndFrom()
```
catstavi, line 248: looks like the contents of `readAnaAndFrom` could be collapsed into this function.

catstavi:
so services should only bump up to this one once everything is on the currently-checked-in version, yeah?

chitang:
No, only soql-postgres-adapter needs to be in the last version 2.11.28.

PR #150, opened 2020-07-14
EN-41091: Fix window function refactor bug that affects rollup.

chitang:
EN-40991: Fix copy function to provide default parameter (#147)
EN-40991: Fix missing replaceRefs in window function info (#148)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "2.12.2"
```
catstavi, line 1: ??

soql-analyzer/src/main/scala/com/socrata/soql/aggregates/AggregateChecker.scala
```diff
@@ -65,6 +67,9 @@ class AggregateChecker[Type] {
   def checkPostgroupExpression(clause: String, e: Expr, groupExpressions: Iterable[Expr]) {
     if(!isGroupExpression(e, groupExpressions)) {
       e match {
+        case FunctionCall(function, params, _) if function.name == SpecialFunctions.WindowFunctionOver =>
+          // Skip the first parameter which is supposedly an aggregate function.
```
catstavi, line 71: first parameter vs. use of 'tail' ?

PR #153, opened 2020-07-14
EN-41091: Fix window function refactor bug that affects rollup.

catstavi:
EN-40991: Fix copy function to provide default parameter (#147)
EN-40991: Fix missing replaceRefs in window function info (#148)

### DIFF
--- a/soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
+++ b/soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
@@ -567,4 +567,10 @@ SELECT visits, @x2.zx
     analysis.selection.head._1.name must equal("sum_a1_id")
   }
 
+
+  test("Merge window function with group by aggregate") {
+    val analysis1 = analyzer.analyzeFullQuery("SELECT name_first, name_last, sum(visits) as n2 GROUP BY name_first, name_last |> SELECT name_first, row_number() over(order by n2) as rn")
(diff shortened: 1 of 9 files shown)


soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
```diff
@@ -567,4 +567,10 @@ SELECT visits, @x2.zx
     analysis.selection.head._1.name must equal("sum_a1_id")
   }
 
+
+  test("Merge window function with group by aggregate") {
```
catstavi, line 571: are these two selects supposed to be equivalent? it's not sure right off that they are

chitang:
I am not sure and will check.  It is funny if they aren't.  We have been merging this type of queries all along.

soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
chitang: Good news, they seem to be identical - at least in blist_dev

SELECT view_type, row_number() over(order by sum(id)) as rn from lenses GROUP BY view_type, display_type 

vs

    SELECT view_type, row_number() over(order by n2) as rn from (
    SELECT view_type, display_type, sum(id) as n2  from lenses GROUP BY view_type, display_type) t1

chitang: btw, this has been this way all long.  it was broken shortly by window function.  it was fixed and this new test was added.

PR #155, opened 2020-07-16
Add SoQL documentation to function spec

rozap:
* Add docstring and examples to Function
* This allows us to define documentation next
  to the declaration, which will make its way
  out to the SoQL Query editor and be surfaced
  on the page in which the user is writing queries
* Actual documentation to be filled in later


soql-analyzer/src/main/scala/com/socrata/soql/functions/Function.scala
```diff
@@ -24,7 +27,9 @@ case class Function[Type](identity: String,
                           parameters: Seq[TypeLike[Type]],
                           repeated: Seq[TypeLike[Type]],
                           result: TypeLike[Type],
-                          isAggregate: Boolean = false)  {
+                          isAggregate: Boolean,
```
catstavi, line 30: was losing the default value here on purpose?

rozap: yep, you can't have default values and variadic arguments in the same signature.

PR #157, opened 2020-07-22
EN-41224: Add extract isoyear

QMalcolm:
Add SoQL function to extract the isoyear from a floating timestamp.

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -450,9 +450,15 @@ object SoQLFunctions {
   val FloatingTimeStampExtractDow = mf("floating timestamp extract day of week", FunctionName("date_extract_dow"), Seq(SoQLFloatingTimestamp), Seq.empty, SoQLNumber)(
     NoDocs
   )
+  // Extracting the week from a floating timestamp extracts the iso week (1-53).
+  // Sometimes the last few days of December may be considered the first week of
+  // next year. See https://en.wikipedia.org/wiki/ISO_week_date for more info.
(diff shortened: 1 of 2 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -453,6 +453,9 @@ object SoQLFunctions {
   val FloatingTimeStampExtractWoy = mf("floating timestamp extract week of year", FunctionName("date_extract_woy"), Seq(SoQLFloatingTimestamp), Seq.empty, SoQLNumber)(
```
znep, line 453: I might change the string to be something like `floating timestamp extract ISO week of year` and/or add a comment that `woy` is ISO week just to leave a few crumbs to make that more obvious to folks in the future despite the function not being named `iso`.

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -453,6 +453,9 @@ object SoQLFunctions {
...
     NoDocs
   )
+  val FloatingTimestampExtractIsoY = mf("floating timestamp extract isoyear", FunctionName("date_extract_iso_y"), Seq(SoQLFloatingTimestamp), Seq.empty, SoQLNumber)(
+    NoDocs
```
rozap, line 457: `NoDocs` :(

If you write a doc, it'll show up in the new query editor :)

QMalcolm: Where should I write said doc? Is there an example?

QMalcolm: The discussion happened elsewhere. There is another branch which is doing a bunch of the initial doc write-ups (https://github.com/socrata-platform/soql-reference/tree/cduranti/add-docs). Once that branch is merged, I'll loop back around making a new pr to add the doc for this new function.

PR #158, opened 2020-08-03
Un-reserve certain reserved words

rjmac:
* Using the postgresql reserved-words list as a guideline:
   https://www.postgresql.org/docs/12/sql-keywords-appendix.html
* As a side effect, "NULLS FIRST" / "NULLS LAST" is now accepted.
  * so is "NULL FIRST" / "NULL LAST" because backwards compat


chitang:
Can we add a test?  Otherwise, lgtm

rjmac:
We've got tests for the existing behavior.  Added one for the new too, though as there's now nothing that marks out an unreserved keyword as a keyword, logically speaking, it's pretty trivial.

chitang:
:shipit:

PR #159, opened 2020-08-14
EN-41671: Add LEAST and GREATEST functions

rjmac:


catstavi:
does this need a version bump?

rjmac:
It does, but that's traditionally a separate commit from the feature itself.

PR #161, opened 2020-09-02
EN-24494: cannot merge chained soql if there is search and grouping a…

chitang:
…ny more after the new semantics of search (application to the output schema instead of the input schema)

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
@@ -575,7 +575,7 @@ private class Merger[T](andFunction: MonomorphicFunction[T]) {
   // for a query that contains a search.
   private def tryMerge(a: Analysis, b: Analysis): Option[Analysis] = (a, b) match {
     case (a, _) if (hasWindowFunction(a)) => None
-    case (SoQLAnalysis(aIsGroup, false, aSelect, Nil, aWhere, aGroup, aHaving, aOrder, aLim, aOff, aSearch),
+    case (SoQLAnalysis(aIsGroup, false, aSelect, Nil, aWhere, aGroup, aHaving, aOrder, aLim, aOff, None),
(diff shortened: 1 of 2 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "2.16.0"
+version in ThisBuild := "2.16.1-SNAPSHOT"
```
langlk, line 1: Is this supposed to say `SNAPSHOT` or should is just by `2.16.1`?

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -575,7 +575,7 @@ private class Merger[T](andFunction: MonomorphicFunction[T]) {
   // for a query that contains a search.
   private def tryMerge(a: Analysis, b: Analysis): Option[Analysis] = (a, b) match {
     case (a, _) if (hasWindowFunction(a)) => None
-    case (SoQLAnalysis(aIsGroup, false, aSelect, Nil, aWhere, aGroup, aHaving, aOrder, aLim, aOff, aSearch),
+    case (SoQLAnalysis(aIsGroup, false, aSelect, Nil, aWhere, aGroup, aHaving, aOrder, aLim, aOff, None),
```
langlk, line 578: Are we adding the search to the query elsewhere? It looks like all of these just remove it, I expect I'm missing something though.

version.sbt
chitang: snapshot now and remove snapshot in two steps is proper.  We skip a lot these days.

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
chitang: These search cases just won't match any more and go to the catch all case which means not mergeable.  Then the chained soql retains it's original form with search.

chitang: These search cases just won't match any more and go to the catch all case which means not mergeable.  Then the chained soql retains it's original form with search.

langlk (approved):

PR #164, opened 2020-10-27
EN-42634: count expression/* + window function

chitang:
Count distinct still does not work with window function.  It is not supported in Postgres (at least in v11) anyway.

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
@@ -247,7 +247,9 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
           sb <- parameters(1).format(d, sb, limit)
         } yield sb.append("]")
       case SpecialFunctions.StarFunc(f) =>
-        Some(sb.append(f).append("(*)"))
+        sb.append(f).append("(*)")
+        window.foreach(_.format(d, sb, limit))
+        Some(sb)
       case SpecialFunctions.Operator(op) if parameters.size == 1 =>
         op match {
           case "NOT" =>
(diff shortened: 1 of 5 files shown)


rjmac (approved):

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -247,7 +247,9 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
           sb <- parameters(1).format(d, sb, limit)
         } yield sb.append("]")
       case SpecialFunctions.StarFunc(f) =>
-        Some(sb.append(f).append("(*)"))
+        sb.append(f).append("(*)")
+        window.foreach(_.format(d, sb, limit))
```
rjmac, line 251: calling `format` can fail if the limit is non-None; you need to check its result (e.g., the way the Subscript function in the previous stanza does)

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -247,9 +247,11 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
...
         } yield sb.append("]")
       case SpecialFunctions.StarFunc(f) =>
-        sb.append(f).append("(*)")
-        window.foreach(_.format(d, sb, limit))
-        Some(sb)
+       sb.append(f).append("(*)")
+       for {
+         w <- window
+         sb <- w.format(d, sb, limit)
+       } yield sb
```
rjmac, line 254: Yes, I mean like that.  Or just `window.flatMap(_.format(d, sb, limit))` would work too, but this is fine.

rjmac (approved):

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -248,10 +248,7 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
         } yield sb.append("]")
       case SpecialFunctions.StarFunc(f) =>
        sb.append(f).append("(*)")
-       for {
-         w <- window
-         sb <- w.format(d, sb, limit)
-       } yield sb
+       window.flatMap(_.format(d, sb, limit)).orElse(Some(sb))
```
rjmac, line 251: Hm no, not like this, actually.... I gave bad advice, sorry :(

More like
```scala
window match {
  case Some(w) =>
    w.format(d, sb, limit)
  case None =>
    Some(sb)
}
```

PR #169, opened 2021-01-31
union

chitang:
Anlaysis

Third approach - binary tree

upgrade analysis from from string to table name

Serializer version

Update serializer

Rightmost

more for rollup

It is previous instead of rightmost

Fix Seq to binary

Must be depth first

Previous

Soqlmerge

Rename function

Fixed tree previous

Further subclass compound

Fix serialization test

Soql merge

Typo PIPEQUERY

Unions in joins

Change join analysis

join+union remapAnalysis

Join - convert from to default _

Handle select *

Fix AliasAnalysis - join (select d1.field from @dog as d1)

Serializer missing something

Scala 2.10 compatibility - Either does not have map until 2.12

Join toString

Column name mapper

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
@@ -113,4 +113,6 @@ case class MinusQuery[T](left: BinaryTree[T], right: BinaryTree[T]) extends Comp
   val op = Compound.Minus
 }
 
-case class Leaf[T](leaf: T) extends BinaryTree[T]
+case class Leaf[T](leaf: T) extends BinaryTree[T] {
+  override def toString: String = leaf.toString
+}
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
@@ -31,13 +31,19 @@ trait BinaryTree[T] {
 }
 
 object Compound {
+  val QueryPipe = "QUERYPIPE"
+  val Union = "UNION"
+  val UnionAll = "UNION ALL"
+  val Intersect = "INTERSECT"
+  val Minus = "MINUS"
+
   def apply[T](op: String, left: BinaryTree[T], right: BinaryTree[T]): Compound[T] = {
     op match {
-      case "QUERYPIPE" => PipeQuery(left, right)
-      case "UNION" => UnionQuery(left, right)
-      case "UNION ALL" => UnionAllQuery(left, right)
-      case "INTERSECT" => IntersectQuery(left, right)
-      case "MINUS" => MinusQuery(left, right)
+      case QueryPipe => PipeQuery(left, right)
+      case Union => UnionQuery(left, right)
+      case UnionAll => UnionAllQuery(left, right)
+      case Intersect => IntersectQuery(left, right)
+      case Minus => MinusQuery(left, right)
     }
   }
 
@@ -46,7 +52,9 @@ object Compound {
   }
 }
 
-abstract class Compound[T](val op: String) extends BinaryTree[T] {
+sealed trait Compound[T] extends BinaryTree[T] {
+
+  val op: String
 
   val left: BinaryTree[T]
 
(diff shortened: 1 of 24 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
```diff
@@ -0,0 +1,90 @@
+package com.socrata.soql
+
+trait BinaryTree[+T] {
```
rjmac, line 3: This should be `sealed` so that the compiler can do exhaustivity checking (and I know it's extended outside this file, and also that there are matches in here which are non-exhaustive as a result -- more specifically, I a pre-analysis tree and post-analysis tree shouldn't be the same type)

rjmac: ..though parameterizing over the leaf type would deal nicely with that.

soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  def last: T = asLeaf
+
+  def asLeaf: T = {
+    this match {
+      case _: Compound[T] => ???
+      case t: T @unchecked => t
+    }
+  }
```
rjmac, line 23: Ugh no.  Partial functions are bad, this should return an `Option[T]` so if it's not a leaf use-sites are forced to deal with that possibilty.

rjmac: (or better still not exist at all and just require use-sites to pattern-match out the leaves)

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -3,38 +3,50 @@ package com.socrata.soql.ast
...
+  *   "join 4x4 as a" => JoinSelect(Left(TableName(4x4, a))) { aliasOpt = a }
   *   "join (select id from 4x4) as a" =>
-  *     JoinSelect(TableName(4x4, None), Some(SubSelect(List(_select_id_), a))) { aliasOpt = a }
+  *     JoinSelect(Right(SubSelect(List(_select_id_), a))) { aliasOpt = a }
   *   "join (select c.id from 4x4 as c) as a" =>
-  *     JoinSelect(TableName(4x4, Some(c)), Some(SubSelect(List(_select_id_), a))) { aliasOpt = a }
+  *     JoinSelect(Right(SubSelect(List(_select_id_), a))) { aliasOpt = a }
   */
-case class JoinSelect(fromTable: TableName, subSelect: Option[SubSelect]) {
+case class JoinSelect(subSelect: Either[TableName, SubSelect]) {
```
rjmac, line 25: `case class(just an Either)` feels weird to me; I think I'd prefer breaking it into a two-case hierarchy, but I donlt feel super-strongly on the subject.

chitang: JoinSelect is reduced to just an Either.  Keeping it just make it a little easier in transition.  Can refactor in the future.

soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
```diff
@@ -0,0 +1,90 @@
...
+trait BinaryTree[+T] {
+
+  def seq: Seq[T] = {
+    Seq(asLeaf)
+  }
+
+  /**
+   * output schema
+   */
+  def previous: T = {
```
rjmac, line 12: This naming seems weird.

rjmac:
Broadly this looks really good to me!  I have some quibbles about the mechanics of Scala, but the whole approach seems pretty clean.  I'd like to see that it does Sensible Things in error cases too.

soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
chitang: any suggestion, outputSchemaLeaf?

rjmac: Sure, that sounds good!  Any hint in the name that the purpose of the method is to say "what is the ultimate schema that this subtree will produce".

rjmac:
..I wonder how much things would be simplified if the analyzer produced a fully-`from`'d analysis instead of keeping the ambient dataset context ambient.  That would be a fairly major refactor on its own but I bet it'd be worth it in the long run now that we're doing nontrivial things with combining queries.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -74,20 +101,21 @@ object Select {
...
   offset: Option[BigInt],
-  search: Option[String]) {
+  search: Option[String]) extends BinaryTree[Select] {
 
   private def toString(from: Option[TableName]): String = {
     if(AST.pretty) {
       val distinctStr = if (distinct) "DISTINCT " else ""
       val selectStr = Some(s"SELECT $distinctStr$selection")
-      val fromStr = from.map(t => s"FROM $t")
+      val fromStr = this.from.orElse(from).map(t => s"FROM $t")
```
saraid, line 118: This `this` seems kinda unnecessary?

chitang: this is this class from.  It is different than the parameter from.  This class from take precedence.  We actually want to remove the parameter from eventually but not yet.

soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
chitang: How do I seal the trait with BinaryTree[Select] and BinaryTree[Analysis] being in a different file/package?
I don't get how case class Leaf helps.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
saraid: Ah. In that case, carry on. Also it's horrifying.

soql-standalone-parser/src/main/scala/com/socrata/soql/BinaryTree.scala
rjmac: You do
```sealed trait BinaryTree[T] {
   ...
}
sealed trait Compound[T] extends BinaryTree[T] {
}
case class /*all the various different types of Compound */ extends Compound[T]
case class Leaf[T](leaf: T) extends BinaryTree[T]
```

Basically anything abstract gets sealed so that a pattern match can absolutely say "I know for a fact that all possible cases of the BinaryTree[T] are covered and there isn't some random `class Bleh extends BinaryTree[Select]` out in some file I don't know about which isn't being handled here."

rjmac (approved):

PR #180, opened 2021-03-25
Fix serialization gap problem

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
```diff
@@ -326,7 +326,7 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
         val seq: NonEmptySeq[SoQLAnalysis[C, T]] = deserializer.read()
         val bt: BinaryTree[SoQLAnalysis[C, T]] = toBinaryTree(seq.seq)
         bt
-      case v if v >= CurrentVersion =>
+      case v if v >= CurrentVersion - 1 =>
```
catstavi, line 329: I might not be understanding this right, but based on my understanding this was an accidental with version 7?

maybe we can make a special exception for the mistake version (7), rather than a continuous exception for all versions one below Current?

also I can see how mistakes happen I don't really get this

chitang: This has changed.  It should be ready for review again.  Everything seems to work although everything is still a mess.

catstavi:
I understand even less now. We don't need any differences in deserialization between LastVersion and CurrentVersion??

chitang:
The difference between last version and current version 7 and 8 are in the details, not in the constructor.

catstavi:
Is it expected that everytime someone adjusts the values of LastVersion and CurrentVersion they also need to adjust how this case statement works?

chitang:
We don't really try to maintain permanent compatibility among all versions for long period of time.  We just want to get a smooth deploy.  I do keep a few things around longer in the jump from Seq to BinaryTree upgrade.  That is also a special change unlike most other change.  That makes things a little extra ugly.

chitang:
There are many ways to get all the puzzle fits.  I don't think there is one single established way especially around BinaryTree.

chitang:
I was skunked because I thought there is a standard way I can casually follow.

catstavi:
Maybe you can leave a comment near current version that the next person who bumps the version needs to adjust the case to differentiate between LastVersion and CurrentVersion?

chitang:
Every change is kind of unique.  I am not sure that comment will be helpful.  I do not know how I want to play the puzzle until I know what I want to do next time and what the world is at that next time.  No need to worry about this that much.

catstavi (approved):

PR #181, opened 2021-03-29
Add /* .. */-style block comments.

rjmac:
They currently _do not_ nest (which is the same behavior as C

```
> select /* haha */ 1 /* gnu */
Outputs:
_1 .. 1 :: number
None
None
has aggregates: false
```


rjmac:
:shipit: received in slack

PR #182, opened 2021-03-31
EN-44901: Make from (via this alias) context available to lateral join

chitang:
Make from context available to lateral join (previously only contexts from preceding joins are available)


znep:
Is there anywhere reasonable to add test coverage on this?

chitang:
I will add some tests and let you know when it is ready for review again.

langlk (approved):

PR #183, opened 2021-04-14
EN-45612 Add text to json cast

rozap:
* so we can turn something into jsonb
* subscripts work with commit in pg-adapter


saraid (approved):
I don't really understand how this is the only function you need, but okay.

PR #184, opened 2021-04-19
Pivot table

chitang:
Infer column types

curl -v -k "http://localhost:6010/resource/_y2zi-93gu?%24query=select+name,subject,max(score)+as+ss+group+by+name,subject+order+by+name,subject+PIVOT+SELECT+namex,biologyx,chemistryx,physicsx"


catstavi:
@chitang I am so sorry, I completely forgot about this. Is there still ongoing discussion about the syntax we want for pivot tables?

chitang:
Marc is supposed to evaluate this solution.  But I haven't heard from him.  I think there is learning benefit we can get by playing in a full setting.  It isn't something that is difficult to back out or totally entangle with the original code base.

PR #185, opened 2021-04-26
Add a `single_row` predefined alias

rjmac:
This is so that you can create virtual tables containing a single row
(e.g., to force single evaluation of expressions) in a manner like this

```
select x, y, z
 join (select something_expensive() as a from @single_row) as vars on true
 where x > @vars.a
```
and ultimately get generated SQL that looks something like
```
select x, y, z
  join (select something_expensive() as a) vars -- note no FROM clause here
  where x > vars.a
```
This is a bit of a workaround for the fact that a SoQL expression
with no `FROM` clause means "select from the current ambient table",
contrary to SQL.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
@@ -349,6 +349,8 @@ class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type],
 
   private def aliasContext(tableName: Option[TableName], ctx: AnalysisContext): AnalysisContext = {
     tableName match {
+      case Some(TableName(TableName.SingleRow, alias)) =>
+        Map(alias.getOrElse(TableName.SingleRow) -> DatasetContext.empty[Type])
       case Some(TableName(name, Some(alias))) =>
         val name1 = if (name == TableName.This) TableName.PrimaryTable.qualifier else name
         Map(alias -> ctx(name1))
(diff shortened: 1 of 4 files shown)


saraid (approved):
Don't know if this catches all cases or not, but the present code makes sense.

soql-environment/src/main/scala/com/socrata/soql/environment/TableName.scala
```diff
@@ -29,6 +29,7 @@ object TableName {
   val SoqlPrefix = "@"
   val PrimaryTable = TableName(SodaFountainPrefix)
   val This = SodaFountainPrefix + "this"
+  val SingleRow = SodaFountainPrefix + "single_row"
```
saraid, line 32: Thought: Also do `val ReservedNames = [This, SingleRow]`, and then we'd be able to throw exceptions if someone tries to create an AnalysisContext using these?

rjmac: Probably, yeah, actually....

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -350,8 +361,10 @@ class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type],
   private def aliasContext(tableName: Option[TableName], ctx: AnalysisContext): AnalysisContext = {
     tableName match {
       case Some(TableName(TableName.SingleRow, alias)) =>
+        alias.foreach(validateAlias(_))
```
saraid, line 364: Hahah oh god someone could have done `FROM @single_row AS this` and I hate that and thank you for nipping that before it budded.

PR #188, opened 2021-05-17
Support for UDFs

rjmac:
* Introduce a "hole" AST node which represents a parameter
* Allow JoinSelect to express join-to-function
* Keep From-qualifiers around when analyzing (fixes a bug in SQL generation)

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
@@ -44,7 +44,7 @@ case class JoinQuery(selects: BinaryTree[Select], definiteAlias: String) extends
 case class JoinFunc(tableName: TableName, params: Seq[Expression])(val position: Position) extends JoinSelect {
   val alias = tableName.alias
   override def toString =
-    TableName.withSoqlPrefix(tableName.name) + "(" + params.mkString(", ") + ") AS " + tableName.aliasWithoutPrefix.get
+    TableName.withSoqlPrefix(tableName.name) + "(" + params.mkString(", ") + ")" + tableName.aliasWithoutPrefix.map(" AS " + _)
 
   def directlyReferencedJoinFuncs = Set(tableName.copy(alias = None))
 
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
@@ -78,11 +78,11 @@ case class JoinFunc(tableName: TableName, params: Seq[Expression])(val position:
     }
   }
 
-  def rewrite(udf: UDF, params: Seq[Expression], aliasProvider: AliasProvider): JoinQuery = {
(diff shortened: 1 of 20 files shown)


rjmac:
retest this please

rjmac:
hm that's apparently not a thing on this repo... ok, I'll just re-touch that last commit

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -22,33 +19,109 @@ case class SubSelect(selects: BinaryTree[Select], alias: String)
...
+    if(udf.arguments.length != params.length) {
+      throw MismatchedParameterCount(expected = udf.arguments.length, got = params.length)(position)
     }
+    val expr = aliasProvider("_expr")
+    val vars = aliasProvider("_vars")
+    val singleRowFrom = Some(TableName(TableName.SingleRow))
+    val labelledJoin =
+      Select(
+        distinct = false,
+        selection = Selection(Some(StarSelection(Some(expr), Nil)),
```
chitang, line 84: Does this kind of select :*, * already work in this nested setting in many processes downstream all the way to sqlize?  Would it be simpler to expand * here if downstream processes don't handle it?

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -22,33 +19,109 @@ case class SubSelect(selects: BinaryTree[Select], alias: String)
...
+                              Seq(StarSelection(Some(expr), Nil)),
+                              Nil),
+        from = singleRowFrom,
+        joins = Seq(
+          InnerJoin(
```
chitang, line 89: move the two JoinQueries to separate functions.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -22,33 +19,109 @@ case class SubSelect(selects: BinaryTree[Select], alias: String)
...
+    JoinQuery(Leaf(labelledJoin), tableName.alias.getOrElse(tableName.name))
+  }
+}
+
+trait AliasProvider {
+  def apply(base: String): String
+}
+object AliasProvider {
+  implicit val instance = new AliasProvider {
+    def apply(base: String) = base + UUID.randomUUID().toString.replaceAll("-", "_")
```
chitang, line 124: randomUUID seems to generate different soql seen by soda fountain.  It seems that a stable generator might be better and has less chance to have fallouts?

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
rjmac: It seems to, but I wouldn't be super-averse to saying "join functions can't select system cols"

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
rjmac: Nah, breaking up the flow like that makes this sort of thing harder to read

chitang (approved):

PR #194, opened 2021-05-28
Table alias rework

rjmac:
Before: table aliases were declared with `AS identifier` where
"identifier" had to conform to the lexical rules for identifier-ness.

Now: table aliases can still be declared that way, but they can
also be declared with `AS @tablename`.  This has two advantages:

1. It makes declaration reflect use.
2. Lexically it allows 4x4s to be used for table identifiers.

Note that even before, `@tablename` wasn't necessarily always a 4x4,
thanks to resource names.

The upshot of all this is that we can expand `JOIN @some-thng` where
"some-thng" is a derived view or other structure that isn't just a
bare table into `JOIN (select ...) AS @some-thng` and have it work.

Also this makes the alias on table-functions optional, as we can use
the function-name as the alias reliably now.

### DIFF
--- a/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
+++ b/soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
@@ -42,7 +42,7 @@ case class JoinTable(tableName: TableName) extends JoinSelect {
 }
 case class JoinQuery(selects: BinaryTree[Select], definiteAlias: String) extends JoinSelect {
   val alias = Some(definiteAlias)
-  override def toString = "(" + Select.toString(selects) + ") AS @" + TableName.removeValidPrefix(definiteAlias) + ""
+  override def toString = "(" + Select.toString(selects) + ") AS @" + TableName.removeValidPrefix(definiteAlias)
   def replaceHoles(f: Hole => Expression): JoinQuery =
     copy(Select.walkTreeReplacingHoles(selects, f))
   def rewriteJoinFuncs(f: Map[TableName, UDF], aliasProvider: AliasProvider): JoinQuery =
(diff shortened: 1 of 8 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/AbstractParser.scala
```diff
@@ -224,9 +224,9 @@ abstract class AbstractParser(parameters: AbstractParser.Parameters = AbstractPa
     if(allowJoinFunctions) {
       // I'd like this alias to be optional, but it would take a
       // surprisingly major change to get that working.
```
saraid, line 226: Don't need this comment anymore!

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -42,7 +42,7 @@ case class JoinTable(tableName: TableName) extends JoinSelect {
 }
 case class JoinQuery(selects: BinaryTree[Select], definiteAlias: String) extends JoinSelect {
   val alias = Some(definiteAlias)
-  override def toString = "(" + Select.toString(selects) + ") AS `" + TableName.removeValidPrefix(definiteAlias) + "`"
+  override def toString = "(" + Select.toString(selects) + ") AS @" + TableName.removeValidPrefix(definiteAlias) + ""
```
saraid, line 45: Can't decide if that ending empty string is mischief or necessary for some mysterious reason.

saraid (approved):
Left 2 comments which amount to nitpicks. This looks fine to me.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
rjmac: Mischief!

PR #195, opened 2021-06-04
add needsWindow to window functions

saraid:
Advice about how to scala better is also welcome.


soql-analyzer/src/main/scala/com/socrata/soql/typechecker/Typechecker.scala
```diff
@@ -104,6 +104,10 @@ class Typechecker[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Typ
       case Right(es) => es
     }
 
+    if(functionInfo.windowFunctions(name) && window.isEmpty) {
```
rjmac, line 107: how about `!functionInfo.windowFunctions(name) && !window.isEmpty`, too?

saraid: Done.

rjmac (approved):

PR #197, opened 2021-06-30
EN-46931: Fix window function check

chitang:
Aggregate functions can have window info


soql-analyzer/src/main/scala/com/socrata/soql/typechecker/Typechecker.scala
```diff
@@ -161,6 +158,19 @@ class Typechecker[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Typ
     }
   }
 
+  private def checkWindowFunction(fc: FunctionCall, candidateFn: MonomorphicFunction[_]): Unit = {
```
saraid, line 161: Is this always going to be a `MonomorphicFunction`? Why not `Function` here?

chitang: We are hitch-hiking on the resolved funtions which are always Monomorphic.  We don't need another way to figure out the Functions(s).

saraid: Good enough for me.

saraid (approved):
Seems to cover the relevant cases.

saraid (approved):
Yeah, that's eleganter.

PR #203, opened 2021-09-14
Be able to preserve parentheses in compound queries

chitang:
Previously, parentheses can only influence the left, right structure of binary tree.  This doesn't work well with set operations that are different in precedence, for example:

(SELECT a FROM a UNION SELECT b FROM b) INTERSECT SELECT c FROM c

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
@@ -27,7 +27,7 @@ private trait DeserializationDictionary[C, T] {
 }
 
 object AnalysisDeserializer {
-  val CurrentVersion = 8
+  val CurrentVersion = 9
   val LastVersion = 7
   val NonEmptySeqVersion = 6
 
@@ -167,12 +167,18 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
     def readBinaryTree[A](f: => A): BinaryTree[A] = {
       in.readUInt32() match {
         case 1 =>
-          Leaf(f)
+          val inParen =
+            if (version >= CurrentVersion || version == TestVersionV5) in.readBool()
+            else false
+          Leaf(f, inParen)
         case 2 =>
           val op = in.readString()
+          val inParen =
+            if (version >= CurrentVersion || version == TestVersionV5) in.readBool()
+            else false
           val l = readBinaryTree(f)
           val r = readBinaryTree(f)
-          Compound(op, l, r)
+          Compound(op, l, r, inParen)
       }
     }
 
@@ -189,7 +195,7 @@ class AnalysisDeserializer[C, T](columnDeserializer: String => C, typeDeserializ
       readSeq {
         val joinType = JoinType(in.readString())
         val joinAnalysis = readJoinAnalysis()
-        val lateral = if (version >= CurrentVersion || version == TestVersionV5) in.readBool() else false
+        val lateral = if (version >= 8 || version == TestVersionV5) in.readBool() else false
         Join(joinType, joinAnalysis, readExpr(), lateral)
       }
     }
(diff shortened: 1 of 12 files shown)


soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -42,11 +42,23 @@ case class JoinTable(tableName: TableName) extends JoinSelect {
 }
 case class JoinQuery(selects: BinaryTree[Select], definiteAlias: String) extends JoinSelect {
   val alias = Some(definiteAlias)
-  override def toString = "(" + Select.toString(selects) + ") AS @" + TableName.removeValidPrefix(definiteAlias)
+
+  override def toString = {
+    if (selects.inParen) {
+      Select.toString(selects) + " AS @" + TableName.removeValidPrefix(definiteAlias)
```
sjzavaleta, line 48: This looks switched around too

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -42,11 +42,23 @@ case class JoinTable(tableName: TableName) extends JoinSelect {
...
+  }
+
   def replaceHoles(f: Hole => Expression): JoinQuery =
     copy(Select.walkTreeReplacingHoles(selects, f))
-  def rewriteJoinFuncs(f: Map[TableName, UDF], aliasProvider: AliasProvider): JoinQuery =
-    copy(selects = Select.rewriteJoinFuncs(selects, f)(aliasProvider))
+
+  def rewriteJoinFuncs(f: Map[TableName, UDF], aliasProvider: AliasProvider): JoinQuery = {
+    val rewritten = Select.rewriteJoinFuncs(selects, f)(aliasProvider)
+    copy(selects = rewritten.wrapInParen)
```
sjzavaleta, line 59: Most changes indeed just preserve parens, but we're also consciously adding them to join functions sometime? When are they rewritten like this?

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -578,7 +578,8 @@ case class JoinAnalysis[ColumnId, Type](subAnalysis: Either[TableName, SubAnalys
     val (subAnasStr, aliasStrOpt) = subAnalysis match {
       case Right(SubAnalysis(subAnalysis, subAlias)) =>
         val selectStr = subAnalysis.toString
-        (s"($selectStr)", Some(subAlias))
+        if (subAnalysis.inParen) (s"$selectStr", Some(subAlias))
```
sjzavaleta, line 581: This looks switched around

soql-analyzer/src/main/scala/com/socrata/soql/AnalysisDeserializer.scala
```diff
@@ -27,7 +27,7 @@ private trait DeserializationDictionary[C, T] {
 }
 
 object AnalysisDeserializer {
-  val CurrentVersion = 8
+  val CurrentVersion = 9
   val LastVersion = 7
```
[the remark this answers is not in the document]
chitang: Not really.  It works as is.  But I streamlined it a bit.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
chitang: It is by design.  JoinQuery is on top of (not part of) the binaryTree.  If the binaryTree has paren, JoinQuery doesn't need another paren.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
chitang: i think we always attempt to rewrite joins.  They are actually rewritten when they are UDFs.  Parens are always safe and necessary for query joins.

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
chitang: Same explanation, it is something on top of the binaryTree.

sjzavaleta (approved):

PR #207, opened 2021-10-14
Add support for scala 2.13, drop support for scala 2.11

rjmac:
Also vastly speed up the parser, and improve the error messages generated by it.


sjzavaleta (approved):
I can't meaningfully review the parser code.  For both, I don't have a good idea of the real rules in my head to compare them against. For the abstract combinator, there's too many unfamiliar operators. Please have someone else look at these

soql-analyzer/src/main/scala/com/socrata/soql/typechecker/Typechecker.scala
```diff
@@ -64,12 +64,12 @@ class Typechecker[Type](typeInfo: TypeInfo[Type], functionInfo: FunctionInfo[Typ
       // Here's where we find that that corresponds to a field access.
       // If we don't find anything that works, we pretend that we
       // never did this and just typecheck it as a subscript access.
-      typecheck(base, aliases, from).right.flatMap { basePossibilities =>
+      typecheck(base, aliases, from).flatMap { basePossibilities =>
```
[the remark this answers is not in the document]
rjmac: Yep, `Either` pre-2.12 was unbiased - `Left` and `Right` were exact mirrors of each other and so if you wanted to operate on one of them you had to project it with the `left` or `right` method.  This is contrary to how `Either` is used in most languages that have it (though I personally rather liked it) where `Right` has the connotation of a success value of some kind, so in 2.12 `Either` was biased `Right`wards and in 2.13 the `right` method was deprecated.

rjmac:
> I can't meaningfully review the parser code. For both, I don't have a good idea of the real rules in my head to compare them against. For the abstract combinator, there's too many unfamiliar operators. Please have someone else look at these

There aren't actually that many operators. ;)  `a ~ b` means "parse an `a` followed by a `b` and return both results", `a | b` means "parse either an `a` or a `b` and return the result of whichever was successful", and `a ^^ f` means "parse an `a`, and if it's successful, transform its result by applying `f` to it."  `~>` and `<~` are like `~` but drop the result of the parser on the side it's pointing away from.

soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -349,20 +349,20 @@ class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type],
 
   def analyzeInOuterSelectionContext(initialCtx: AnalysisContext)(lastQuery: Analysis, query: Select): Analysis = {
     val prevCtx = contextFromAnalysis(lastQuery)
-    val nextCtx = prevCtx ++ initialCtx.filterKeys(qualifier => TableName.PrimaryTable.qualifier != qualifier)
+    val nextCtx = prevCtx ++ initialCtx.view.filterKeys(qualifier => TableName.PrimaryTable.qualifier != qualifier)
```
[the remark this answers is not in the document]
rjmac: In 2.12 and previous, `filterKeys` was a lazy operation that implicitly returned a view on the original map rather than a copy of the map with the filtered keys removed.  In 2.13 it was made explicit (and I added the `filteredKey` method to `IterableView`-of-`Map` in the `com.socrata.soql.collection` package object; ditto `mapValues` which also created an implicit view before)

soql-environment/src/main/scala/com/socrata/soql/environment/DatasetContext.scala
```diff
@@ -10,7 +10,11 @@ trait UntypedDatasetContext {
 trait DatasetContext[Type] extends UntypedDatasetContext {
   override protected implicit def selfContext = this
   val schema: OrderedMap[ColumnName, Type] // Note: contains ALL columns, system AND user!
-  lazy val columns: OrderedSet[ColumnName] = schema.keySet
+  lazy val columns: OrderedSet[ColumnName] = {
+    val result = OrderedSet.newBuilder[ColumnName]
+    result ++= schema.keys
+    result.result()
+  }
```
saraid, line 17: That just makes me sad.

rjmac: Indeed.

soql-standalone-parser/src/test/scala/com/socrata/soql/parsing/ParserTest.scala
```diff
@@ -284,7 +284,7 @@ class ParserTest extends WordSpec with MustMatchers {
         fail("Unexpected success")
       } catch {
         case e: BadParse =>
-          e.message must equal (parser.errors.leafQueryOnTheRightExpected)
+          e.message must equal ("Expected a non-compound query, but got `('")
```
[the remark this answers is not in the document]
saraid: "Expected non-compound query on right-hand side of pipe operator"?

soql-standalone-parser/src/test/scala/com/socrata/soql/parsing/HandParseTest.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  test("combinators and hand parser match - statements") {
+    def go(s: String): Unit = {
+      val a = new HRParser()
+      val b = new StandaloneParser()
+      if(printTimings) println(s)
+      timing("hand-rolled") { a.unchainedSelectStatement(s) } must equal(timing("combinator") { b.unchainedSelectStatement(s) })
+    }
+
+    go("SELECT 1")
```
saraid, line 82: Reading the actual parser code later, but this seems like a good opportunity to add support for an empty `SELECT`, too.

saraid: ...well, I'm glad I didn't try to dig in more on reading the Actual Parser Code last Friday. -_-

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentParser.scala
```diff
@@ -0,0 +1,1521 @@
...
+            case _ =>
+              reader.addAlternates(TABLEIDSTARSET)
+              reader
+          }
+        case star@STAR() =>
+          val ParseResult(r2, exceptions) = selectExceptions(r.rest, userIdentifier)
+          result += StarSelection(None, exceptions).positionedAt(star.position)
+          loop(r2, commaFirst = true)
+        case _ =>
+          reader.addAlternates(TABLEIDSTARSET)
```
saraid, line 551: This pings weirdly for me, but it's the same as the systemStarSelection bit, so it's probably fine.

Mostly finding it odd that the error message will suggest a table-identifer-plus-star when... not finding one is perfectly okay.

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentParser.scala
```diff
@@ -0,0 +1,1521 @@
...
+      exprResult match {
+        case Some(ParseResult(r2, e)) =>
+          r2.first match {
+            case AS() =>
+              // now we must see an identifier...
+              val ParseResult(r3, identPos) = alias(r2.rest)
+              result += SelectedExpression(e, Some(identPos))
+              loop(r3, commaFirst = true)
+            case _ =>
+              r2.addAlternates(ASSET)
```
saraid, line 609: I understand why it's named `ASSET` and I still hate it.

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentParser.scala
saraid: It _does_ make sense on a node level, and I have the same kind of "wait, this should never be relevant" confusion elsewhere, so I'm inclined to feel that this is a completeness "cover all cases, whether or not they can be walked into naturally because we might need to walk into them from a different direction someday" thing rather than a mistake.

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentParser.scala
```diff
@@ -0,0 +1,1521 @@
...
+  // line-noise.  Leaving them here in case we want to bring them back
+  // for some reason...
+
+  // private val COLONCOLONSET = s(COLONCOLON())
+  // private val MINUSPLUSSET = s(MINUS(), PLUS())
+  // private val CARETSET = s(CARET())
+  // private val FACTORSET = s(STAR(), SLASH(), PERCENT())
+  // private val TERMSET = s(PLUS(), MINUS(), PIPEPIPE())
+  // private val COMPARISONSET = s(EQUALS(), LESSGREATER(), LESSTHAN(), LESSTHANOREQUALS(), GREATERTHAN(), GREATERTHANOREQUALS(), EQUALSEQUALS(), BANGEQUALS())
+  private val OPERATORSET = s(AnOperator)
```
saraid, line 193: I don't know if I agree. Specific operators are relevant in specific cases, and the others are not relevant in those specific cases. Saying "an operator" suggests that any operator would work and that's not true.

But on the other hand, I'm unable to come up with an actual situation in which it doesn't make sense to say "an operator". (If you can, that would be interesting.) (And for the problem of ComparisonSet specifically, if we want it back, I'd suggest collapsing it to "a comparison operator" or some such.)

saraid (approved):
Couldn't find any real problems (`ASSET` is a non-real problem), and no confusions.

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentParser.scala
rjmac: Yeah, I should probably underscore all of the *SET constants...

PR #210, opened 2021-10-20
Rjm/more improvements for dsmapi

rjmac:
- Forgot the DSMAPI extends the expression syntax for let-binding
- so expose a bunch of things that constitute a reasonable API
- and add a handful of useful productions around identifiers
- Also make the QUERYPIPE token print more nicely


vickingan (approved):
👍 LGTM

PR #213, opened 2021-11-02
EN-49069: Less soql merges

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -844,8 +848,28 @@ private class Merger[T](andFunction: MonomorphicFunction[T]) {
...
+        fc.window.nonEmpty || fc.parameters.exists(hasWindowFunction)
+      case _ =>
+        false
+    }
+  }
+
+  private def hasLiteral(e: CoreExpr[_, _]): Boolean = {
+    e match {
+      case fc: com.socrata.soql.typed.FunctionCall[_, _] =>
+        fc.window.nonEmpty || fc.parameters.exists(hasLiteral)
```
rjmac, line 870: Why does "has a window clause" make "hasLiteral" true?

chitang: No it doesn't.  It was a copy paste error.

rjmac (approved):

PR #214, opened 2021-11-05
Add interval datatype

chitang:
Add date_add function

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -61,6 +61,9 @@ object SoQLFunctions {
   val TextToTime = mf("text to time", SpecialFunctions.Cast(SoQLTime.name), Seq(SoQLText), Seq.empty, SoQLTime)(
     NoDocs
   )
+  val TextToInterval = mf("text to interval", SpecialFunctions.Cast(SoQLInterval.name), Seq(SoQLText), Seq.empty, SoQLInterval)(
+    NoDocs
+  )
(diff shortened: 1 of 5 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -568,6 +571,10 @@ object SoQLFunctions {
     NoDocs
   )
 
+  val TimeStampAdd = f("timestamp add", FunctionName("date_add"), Map("a" -> TimestampLike, "b" -> Set(SoQLInterval)), Seq(VariableType("a"), VariableType("b")), Seq.empty, VariableType("a"))(
```
rjmac, line 574: ...`Map("a" -> TimestampLike), Seq(VariableType("a"), FixedType(SoQLInterval))`...

rjmac: Also, any reason this isn't a `SpecialFunctions.Operator("+")` overload?  That should work fine...

rjmac (approved):
Could also add a conversion over in SoQLTypeInfo so that string literals which are formatted correctly are also interval literals, if you liked.

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
chitang: I get what I want without Interval literals.  I will keep that minimal.

chitang: i will try +

PR #216, opened 2021-11-09
Produce better error messages around *-positioning in select lists.

rjmac:
In particular, tell the user that `:*` and `*` must come first, and
that `:*` must come before `*`.


catstavi (approved):
weird that the exceptions have to be defined twice?

rjmac:
Yeah, it's unfortunate. I haven't come up with a good way around it yet, though.  The problem is that the parser should be able to be used standalone, and the analyzer (which depends on the parser, not vice-versa) wants all its exceptions to be unified.  So there's one set of exceptions when the parser is being used by itself, and one when it's being used by the analyzer...

PR #217, opened 2021-11-22
add wrapper for ST_Area, ST_Intersection, and NTILE(N) functionality

PeterAustinMoore:
Adding:
`area` => ST_Area - return the area of a geometry
`polygon_intersection` => ST_Intersection - return the geometry of a an intersecting area (or null if they don't intersect)
`ntile(N)` => NTILE - integer ranging from 1 to the argument value, dividing the partition as equally as possible. Useful for getting quantiles through SoQL.

Bump version

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -167,9 +167,9 @@ object SoQLFunctions {
     Seq(VariableType("a"), VariableType("b")), Seq.empty, FixedType(SoQLBoolean))(
     "Return the rows that where the locations 'spatially overlap', meaning they intersect, but one does not completely contain another and they share interior points")
   
-  val Intersection = f("intersection", FunctionName("intersection"), Map("a" -> Set(SoQLMultiPolygon, SoQLPolygon), "b" -> Set(SoQLMultiPolygon, SoQLPolygon)),
+  val Intersection = f("polygon_intersection", FunctionName("polygon_intersection"), Map("a" -> Set(SoQLMultiPolygon, SoQLPolygon), "b" -> Set(SoQLMultiPolygon, SoQLPolygon)),
     Seq(VariableType("a"), VariableType("b")), Seq.empty, FixedType(SoQLMultiPolygon))(
-    "Returns the geometry of the intersection between two polygon or multipolygon geometries")
+    "Returns the geometry of the overlapping multipolygon intersection between two polygon or multipolygon geometries")
   
   val Area = f("area", FunctionName("area"), Map("a" -> Set(SoQLMultiPolygon, SoQLPolygon)),
     Seq(VariableType("a")), Seq.empty, FixedType(SoQLNumber))(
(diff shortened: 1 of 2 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -166,6 +166,15 @@ object SoQLFunctions {
   val Overlaps = f("overlaps", FunctionName("overlaps"), Map("a" -> GeospatialLike, "b" -> GeospatialLike),
     Seq(VariableType("a"), VariableType("b")), Seq.empty, FixedType(SoQLBoolean))(
     "Return the rows that where the locations 'spatially overlap', meaning they intersect, but one does not completely contain another and they share interior points")
+  
+  val Intersection = f("intersection", FunctionName("intersection"), Map("a" -> GeospatialLike, "b" -> GeospatialLike),
+    Seq(VariableType("a"), VariableType("b")), Seq.empty, VariableType("a"))(
```
rjmac, line 171: This says that an intersection can be taken between any two geometries and will have the same type as the first one.  Is that true?  It seems unlikely.  In general, it seems unlikely that the type of the resulting geometry can be tied to to the type of the input geometries in any particular way (e.g., the intersection of two polygons may be a multipolygon)

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -166,6 +166,15 @@ object SoQLFunctions {
...
+    "Returns the geometry of the intersection between two geometries")
+  
+  val Area = f("area", FunctionName("area"), Map("a" -> GeospatialLike),
+    Seq(VariableType("a")), Seq.empty, FixedType(SoQLNumber))(
+    "Returns the area of the geometry")
```
rjmac, line 176: ..is this useful on non-polygon-like geometry types?

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -499,6 +508,9 @@ object SoQLFunctions {
   val Lag = f("lag", FunctionName("lag"), Map.empty, Seq(VariableType("a")), Seq.empty, VariableType("a"), needsWindow = true)(
     NoDocs
   )
+  val Ntile = f("ntile", FunctionName("ntile"), Map.empty, Seq(VariableType("a")), Seq.empty, FixedType(SoQLNumber), needsWindow=true)(
```
rjmac, line 511: This does not take any random type as its pararameter...

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
rjmac: (also, is this useful on non-polygon-like geometry types?)

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
PeterAustinMoore: technically possible, but less than useful - updated to make them specifically poly/multipolygon types

rjmac (approved):

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
PeterAustinMoore: fair point - I'm going to rename this function such that the "this is between two polygons" and results in a single multipolygon (or null) is clear

PR #219, opened 2021-12-01
EN-49288: Aggregate function call filter

chitang:


soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -293,9 +294,10 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
         }
         (whens ++ otherwise).encloseNesting(d"CASE" flatAlt d"CASE ", d"", d"END" flatAlt d" END")
       case other =>
+        val filterd = filter.fold(Doc.empty) { f => d" FILTER(WHERE ${f.doc})" }
```
rjmac, line 297: This will produce ugly results if it decides to break inside the filter.   Something like

```
(doc"WHERE" +#+ f.doc).enclouseNesting(doc" FILTER(", Doc.empty, ")")
```
would probably be better.

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Expression.scala
```diff
@@ -293,9 +294,10 @@ case class FunctionCall(functionName: FunctionName, parameters: Seq[Expression],
...
         val close = window.fold(Doc.empty) { w => d" " ++ w.doc }
         if(parameters.lengthCompare(1) == 0) {
-          d"${other.toString}(${parameters(0).doc})$close"
+          d"${other.toString}(${parameters(0).doc})$filterd$close"
```
rjmac, line 300: And this might be better as something along the lines of

```
(Seq(Doc(other.toString) ++ d"(" ++ parameters(0).doc ++ d") ++ filter ++ close)).vsep.hanging(2)
```
..in which case you'll want to get rid of the leading spaces on `filter` and `close`.

soql-standalone-parser/src/test/scala/com/socrata/soql/parsing/ParserTest.scala
```diff
@@ -218,7 +218,7 @@ class ParserTest extends WordSpec with MustMatchers {
       println("test 1")
       val x = parseFull("select count(distinct column)")
       x.selection.expressions.head.expression.asInstanceOf[FunctionCall].functionName.name must be ("count_distinct")
-      x.selection.expressions.head.expression.toString must be ("COUNT(DISTINCT `column`)")
+      x.selection.expressions.head.expression.toString must be ("count(DISTINCT `column`)")
```
catstavi, line 221: huh, this is not a thing I would have expected to change?

catstavi (approved):
one question about that test change, otherwise:

I don't see anything wrong here (also yay tests!) but I don't really trust my grasp on this code for an in-depth review. I'll leave it to you if you want more feedback than what I can give!

soql-standalone-parser/src/test/scala/com/socrata/soql/parsing/ParserTest.scala
chitang: It is intentional changed.  You didn't see where? In functioncall.doc

catstavi: oh! yeah I do see it. but why?

chitang: More consistent with function name style.  I was gonna to give reason but forgot.

PR #226, opened 2021-12-27
EN-49813: query hint /*+ hints */

chitang:


rjmac (changes requested):
There's no reason to make these hints like comments.  Choose some other thing, putting semantic data in comments is a curse caused by needing to smuggle information past some other thing that doesn't understand it, and soql isn't in that position.

rjmac (changes requested):
Also I see no validation or interpretation of these hints.  They need to be parsed here, not just blindly passed through.

version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "3.15.0"
+version in ThisBuild := "4.0.0"
```
saraid, line 1: Did we downgrade sbt or something?

rjmac: Huh, I didn't notice that.  That shouldn't have changed back.  :eyes:

rjmac: Actually it doesn't look like this project uses sbt-release at all, so I'm not sure what changed that.

PR #227, opened 2021-12-29
EN-49813: SoQL Hints

chitang:
SELECT
  HINT(materialized, . . .)
  C1, C2, C3 . . .
 WHERE . . .

Vs MSSQL

SELECT . . .
 WHERE . . .
OPTION(hint_1, . . . )

Vs Oracle

SELECT
  /*+ hint_1, . . . */


soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/RecursiveDescentHintParser.scala
```diff
@@ -0,0 +1,54 @@
...
+
+trait RecursiveDescentHintParser { this: RecursiveDescentParser =>
+
+  protected final def hints(reader: Reader): ParseResult[Seq[Hint]] = {
+    reader.first match {
+      case HINT() =>
+        reader.rest.first match {
+          case LPAREN() =>
+            parseHintList(reader.rest.rest)
+        }
```
rjmac, line 18: This will throw a MatchError if it sees a HINT followed by not-an-LPAREN.  You'll want a `case _ => ParseResult(reader, Seq.empty)` here too, to not consume the branch (or `fail` if you do want to keep `hint` a reserved word)

soql-standalone-parser/src/main/scala/com/socrata/soql/parsing/LexerReader.scala
```diff
@@ -105,6 +105,9 @@ object LexerReader {
     INNER,
     LATERAL,
 
+    // HINT
+    HINT,
```
rjmac, line 109: Does this have to be a reserved word, or can it be a keyword?

rjmac (changes requested):
Looks pretty ok, I have some quibbles.

rjmac (approved):

PR #233, opened 2022-01-20
reproject soql function

PeterAustinMoore:
Adds:
A `reproject` SoQL function that loosely does what ST_Transform does in PostGIS, however it is setup to accept only the Proj4 projection text, rather than SRID integers, in order to allow for greater flexibility as well as match what DSMAPI does as a transform: https://dev.socrata.com/docs/transforms/reproject.html

Example: 
```
reproject('POINT (372728.3308536674 69825.24782297359)'::point, '+init=epsg:3627', '+init=epsg:4326')
```

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -698,6 +698,11 @@ object SoQLFunctions {
   val PointToLatitude = field(SoQLPoint, "latitude", SoQLNumber)
   val PointToLongitude = field(SoQLPoint, "longitude", SoQLNumber)
 
+  val Reproject = f("reproject", FunctionName("reproject"), Map("a" -> GeospatialLike), Seq(VariableType("a"), FixedType(SoQLText), FixedType(SoQLText)), Seq.empty, VariableType("a"))(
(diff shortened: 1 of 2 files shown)


chitang (approved):

soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -698,6 +698,11 @@ object SoQLFunctions {
   val PointToLatitude = field(SoQLPoint, "latitude", SoQLNumber)
   val PointToLongitude = field(SoQLPoint, "longitude", SoQLNumber)
 
+  val Reproject = f("reproject", FunctionName("reproject"), Map("a" -> GeospatialLike), Seq(VariableType("a"), FixedType(SoQLText), FixedType(SoQLText)), Seq.empty, VariableType("a"))(
+    "Reproject the incoming geometry from the incoming PROJ.4 spatial reference to an outgoing spatial reference", 
+    Example("Reproject Point", "reproject('POINT (372728.3308536674 69825.24782297359)'::point, '+init=epsg:3627', '+init=epsg:4326')", "")
```
chitang, line 703: what does +init= do?

PeterAustinMoore:
We decided this would not be performant for many operations, encourages bad practices (i.e. not transforming it on ingress), and might cause CPU issues that would negatively affect everyone.

PR #236, opened 2022-02-03
EN-49729: Add user parameter functions and a type for this context

rjmac:


sjzavaleta (approved):
The function names will probably change and we'll probably want friendlier name for dates than e.g. floating_timestamp_parameter but since naming decisions are still in committee, I'd just stick with this for now

PR #241, opened 2022-02-14
EN-50835: Support no_chain_merge query hint

chitang:


rjmac (approved):
`b.hints.exists(_.isInstanceOf[NoChainMerge])` is probably cleaner, but I don't insist upon it.

rjmac (approved):

PR #250, opened 2022-05-09
Fix publish task

rjmac:
It was trying to build with scala 2.12.10, then 2.12.14, and of course
the publish of the latter was failing due to 2.12 artifacts already
existing.  I don't quite understand why - the state SBT was in was
that the root project was set to 2.12.14 from SBT's default and the
subprojects had settings of 2.12.10.  I'm not sure why `+publish` was
taking the union of all of that even for the subprojects.

But now, all the projects use the base `BuildSettings.buildSettings` so
they all agree on the set of scala versions in use.  Also I updated the
build to the newer 2.12 Just Because.


rjmac:
Closed because obsolete, it was also fixed in #265

PR #251, opened 2022-05-31
EN-53559: SoQL distinct on (expr1 [, expr2, ...])

chitang:


soql-analyzer/src/main/scala/com/socrata/soql/SoQLAnalyzer.scala
```diff
@@ -176,6 +178,7 @@ class SoQLAnalyzer[Type](typeInfo: TypeInfo[Type],
     * @throws com.socrata.soql.exceptions.SoQLException if the query is syntactically or semantically erroneous.
     * @return The analysis of the query.  Note this should be appended to `sourceFrom` to form a full query-chain. */
   def analyzeSplitQuery(distinct: Boolean,
+                        distinctOn: Option[String],
```
rjmac, line 181: Does `true, Some(...)` make sense as values here?  If not, some other way to express this would be better (ditto everywhere else distinct + distinctOn exists in parallel)

chitang: No, they cannot be used simultaneously.

chitang: How about distinct: Either[Boolean, Seq[Expression]]?

soql-standalone-parser/src/main/scala/com/socrata/soql/ast/Select.scala
```diff
@@ -423,6 +430,20 @@ case class Select(
...
     docWithFrom(Some(fromTable)).layoutSmart(LayoutOptions(pageWidth = PageWidth.Unbounded)).toString
   }
+
+  def validate(): Select = {
+    distinct match {
+      case DistinctOn(exprs) if orderBys.nonEmpty =>
+        val orderBySet = orderBys.map(_.expression).toSet
+        val distinctNotInOrderBy = (exprs.toSet -- orderBySet)
+        if (distinctNotInOrderBy.nonEmpty) {
+          throw new BadParse("SELECT DISTINCT ON expressions must match initial ORDER BY expressions", distinctNotInOrderBy.head.position)
```
rjmac, line 440: This isn't really a _parse_ error per se but that's fine for now I guess.  But I think is too loose - it should probably be `orderBys.take(exprs.size).map etcetc` to get the "initial" part of the condition right.

rjmac (changes requested):
Just the one quibble, then :shipit:

PR #255, opened 2022-06-13
EN-54117: stop resolving joined (qualified) cols to aliased selections

catstavi:
bug: 'select `name_first` as name_last, @a1.`name_last` as a1_name_last' was resolving a1_name_last as a column ref of `name_first` on the not-joined dataset
this was because we did not take the selection qualifier into account when looking up the column in the alias map.


rjmac (approved):
I fear breaking existing queries, but :yo_yo:

rjmac:
(or worse, changing the meaning of existing queries...)

catstavi:
Tested this locally by bumping soql-postgres-adapter and query-coordinator to my locally published version, and confirmed it fixes the issue. Won't be fixed in Explore Canvas until we can bump the version in dsmapi, which is currently at 4.5.0, which changes I'm not sure how to integrate without learning a lot.

Any other thoughts for testing from folks before I close?

PR #258, opened 2022-08-08
EN-55111: Add more rollup hints

chitang:


rjmac (approved):

rlvoyer:
Please add a description to you PR descriptions to give the team and reviewers some context into what motivated the change.

chitang:
This new rollup application capability is going to require hint to activate.  It is safer and allow us to experiment and learn with less worry (not unlike ff).  This also has other benefits like not forcing us to incur unnecessary overhead in looking up rollups especially when we haven't optimized rollup discovery.

PR #259, opened 2022-08-10
EN-55189: Allow merging with distinct where a.distinctiveness is indistinct and…

chitang:
… b.distinctiveness is indistinct or fullydistinct


chitang:
actually, i want to wrap the 2nd for in a mergeAnalysis function so that there will be very little identation change of the original tryMerge function.

MarkWillisford (approved):

PR #260, opened 2022-08-12
EN-55230: Be able to merge chain query within join sub-analysis

chitang:


soql-analyzer/src/test/scala/com/socrata/soql/SoQLAnalyzerTest.scala
```diff
@@ -765,4 +765,54 @@ SELECT visits, @x2.zx
...
   }
+
+  test("indistinct, distinct is merged") {
+    val analysis1 = analyzer.analyzeFullQueryBinary("SELECT name_first, name_last |> SELECT distinct name_first")
+    val analysis2 = analyzer.analyzeFullQueryBinary("SELECT distinct name_first")
+    val merged = SoQLAnalysis.merge(TestFunctions.And.monomorphic.get, analysis1)
+    merged must equal (analysis2)
+  }
+
+  test("distinct, indistinct should be mergeable but the merger does not support this yet") {
```
catstavi, line 776: sorry so is this a test of behavior that we actually don't want ultimately?

chitang: I don't know about ultimately.  But it is something we don't support currently.

catstavi (approved):

PR #262, opened 2022-09-23
EN-54831 do not automatically insert a dash in generataed aliases

branweb1:
For a new param, we would generate a name like
param_aaaa-aaaa_something, and that would fail column name validation in core since it doesn't allow dashes


branweb1:
i thought about that, but this change had so many pieces that I just put a pin in that thought and made the smallest possible change here so I could move on. Can def do this in more places if it's better.

PR #263, opened 2022-09-26
Revert "EN-55189: Allow merging with distinct where a.distinctiveness…

rjmac:
… is indistinct and… (#259)"

This reverts commit 825fd6570518184b2f5829ff9eaf9085a7f60f73.


saraid (approved):
No version bump? Might also be worth adding a test on the problem case we found.

rjmac:
I generally do version bumps in a separate commit

PR #264, opened 2022-09-28
Remove apparently unnecessary special case in the Typechecker

rjmac:
I have no idea why this clause was here?


sjzavaleta:
And causes issues like https://socrata.atlassian.net/browse/EN-55523

sjzavaleta (approved):

rjmac:
..how did this pass the automated PR checker?  It doesn't actually work.  Reverted on another branch.

PR #265, opened 2022-10-15
Don't build for 2.12 twice (+ update scala versions)

rjmac:


saraid (approved):
I don't understand how this works, but it doesn't look scary.

PR #267, opened 2022-11-29
New analyzer

rjmac:


saraid (approved):
Signing off on this despite not fully wrapping my head around it.

PR #268, opened 2022-12-09
Track expression positions with more nuance

rjmac:
We now have the notion of "logical position" and "physical position". The physical is where an Expr comes from physically within a query's text. If that expr is copied to another position in the text (e.g., via alias expansion or UDF parameter substitution) the logical position is updated to refer to the place the expression has been expanded into.


saraid (approved):
I don't see any problems with this.

PR #272, opened 2022-12-20
EN-57267, EN-57268: Optional params for LAG and LEAD, epoch_seconds

rjmac:
Provide prototypes for LEAD and LAG which include their optional "offset" and "default" parameters, and provide a function for turning a fixed_timestamp into an epoch-seconds number.


saraid (approved):
moar functions for the function god

PR #274, opened 2022-12-20
Refactor various rewrite-pass stuff

rjmac:
Mostly pulling it them out into their own files, and putting them all in the `rewrite` sub-package of analyzer2, but also heavily reworking the preserveOrdering rewrite so that it doesn't need the `row_number` function anymore (which both makes it easier to use and might let PG's query optimizer see what we're doing with this?)

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
@@ -117,6 +117,8 @@ class PreserveOrdering[RNS, CT, CV] private (provider: LabelProvider) {
   }
 }
 
+/** Attempt to preserve ordering from inner queries to outer ones.
+  * SelectListReferences must not be present (this is unchecked!!). */
 object PreserveOrdering {
   def apply[RNS, CT, CV](labelProvider: LabelProvider, stmt: Statement[RNS, CT, CV]): Statement[RNS, CT, CV] = {
     new PreserveOrdering[RNS, CT, CV](labelProvider).rewriteStatement(stmt, true, false)._2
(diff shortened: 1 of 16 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
```diff
@@ -0,0 +1,124 @@
...
+          // make sure we've put the relevant expressins in our
+          // select list and return them.
+          val (newColumns, outputInfo) = orderedSelf.orderBy.map { case OrderBy(expr, asc, nullLast) =>
+            val columnLabel = provider.columnLabel()
+            (columnLabel -> NamedExpr(expr, freshName("order")), (columnLabel, expr.typ, asc, nullLast))
+          }.unzip
+
+          val newSelf = orderedSelf.copy(selectList = selectList ++ newColumns)
+
+          (outputInfo, newSelf)
```
rjmac, line 84: This currently just creates new output columns to reflect the ordering; it could in principle check to see if those exprs are already being selected and just re-use them.  Perhaps worth it...

rjmac:
Grarh did I do a 2.13ism again...?  I did!

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
rjmac: yes, I decided it was worth it.

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
```diff
@@ -0,0 +1,132 @@
...
+
+        // We will at the very least want to add the ordering
+        // columns from the subquery onto the end of our order-by...
+        val orderedSelf = select.copy(
+          from = newFrom,
+          orderBy = orderBy ++ extraOrdering
+        )
+        if(wantOrderingColumns) {
+          // ..and our caller wants to know how we're ordered, so
+          // make sure we've put the relevant expressins in our
```
saraid, line 76:
```suggestion
          // make sure we've put the relevant expressions in our
```

saraid (approved):
I think I followed most of that.

PR #277, opened 2022-12-28
Errors rework

rjmac:
New-analyzer errors are now unified and serializable as JSON

This includes some small other changes:
  * Makes `distinct on`-not-prefix-of-`order by` not a parse error (instead it's an analysis error in the old analyzer)
  * `AbstractLexer` is now public
  * The combinator parser is now gone as it was starting to bitrot
  * The existing LexerParserError hierarchy is now `sealed`


saraid (approved):
LGTM.

PR #284, opened 2023-01-05
Add a paging rewrite to SoQLAnalysis

rjmac:


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/AddLimitOffset.scala
```diff
@@ -0,0 +1,56 @@
...
+
+class AddLimitOffset[RNS, CT, CV] private (labelProvider: LabelProvider) {
+  import AddLimitOffset.zero
+
+  type Statement = analyzer2.Statement[RNS, CT, CV]
+
+  def rewriteStatement(stmt: Statement, desiredLimit: Option[BigInt], desiredOffset: Option[BigInt]): Statement = {
+    stmt match {
+      case CombinedTables(_, _, _) | Values(_) =>
+        // combinedtables and values don't specify an order, so we don't need to preserve an order
```
saraid, line 15: Not sure this comment is relevant?

rjmac: It largely isn't.  It's a leftover from my initial thought about how this would work for selects, which was "add a wrapper `Select` much as this case does, but with an appropriate `order by` before realizing it'd just be way easier to rewrite the existing lim/off.

saraid (approved):
Zero-indexed offsets are wheeeee.

rjmac:
> Zero-indexed offsets are wheeeee.

PR #286, opened 2023-01-13
When pretty-printing strings, use JSON-style rather than SQL-style

rjmac:
Because JSON-style strings with control characters print in a more useful way than SQL-style.


saraid (approved):
Little bit weirded out by this choice, but I'm sure it's fine.

saraid (approved):
JSON say, JSON do.

PR #287, opened 2023-01-30
Rjm/metatypes

rjmac:
Add the concept of a type-universe (a "metatype") which subsumes CT, CV, RNS (+ other stuff that wasn't parameterized before)

This change is to support two things.  The first is cleaning up some ugliness; a _lot_ of things take type parameters to represent types, values, and resource name scopes.  Now they all take the same thing, some subtype of MT, so there's just less physical typing going on.  But this was motivated by wanting to type-parameterize DatabaseTableName and DatabaseColumnName, and _five_ type parameters being lugged around everywhere is just too much.

The reason I wanted to parameterize DTN and DCN is that different layers of the stack have different ideas about what these things are, so they'll need to be rewritten as a query progresses through them (to the point where I _think_ at the SQL-generation level we won't want them to be very stringlike at all), and parameterizing them allows those rewrites to be reflected in the type system.

Additional changes that fell out while I was working on this:
* AutoTableLabel and DatabaseTableName are now completely separate; TableLabel itself is gone entirely.
* Values now uses AutoColumnLabels instead of DatabaseColumnLabels for its columns.  It was always wrong to do the other, the refactory just exposed it.
* TypeInfo has been reworked for a stronger separation between analyzer1's needs from it and analyzer2's.
* When finding datasets, the TableFinder implementation is responsible for generating DatabaseColumnNames rather than being implicit.
* There are new helpers that basicially just make it easier to refer to various types (e.g., the *Universe mixins in MetaTypes.scala, or the projectors in the types subpackage).  These are pure sugar though.


saraid (approved):
Can't find any problems. 95% was just changing the type invocation, which I'd expect the compiler to catch any problems with.

PR #292, opened 2023-02-16
Move analyzer2 serialization into its own subproject

rjmac:
So that it can be implemented for SoQLTypes and SoQLValues

Also move SoQLPosition into soql-environment from the parser in order to break a circular dependency incurred by soql-parser's test dependency on soql-types.


saraid (approved):
I don't see any problems with these changes.

PR #302, opened 2023-03-24
Add a way to recover the view graph from a FoundTables

rjmac:


saraid (approved):
Love how straightforward this is.

PR #304, opened 2023-04-12
"Remove trivial selects" rewrite path

rjmac:
Let's get rid of them in a separate path from the generic "merger" because it's straightforward to do and the logic is much clearer when there isn't a bunch of other stuff going on.


saraid (approved):
Neat.

PR #305, opened 2023-04-13
Generalize "remove trivial selects"

rjmac:
Before it would consider removing a trivial select only if it were the _only_ input dataset to a query.  Now it will do for all trivial-select inputs even in the presence of joins.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/From.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/From.scala
@@ -29,6 +29,8 @@ sealed abstract class From[MT <: MetaTypes] extends LabelUniverse[MT] {
   // into a corner?
   def unique: LazyList[Seq[Column[MT]]]
 
+  def schema: Seq[(AutoTableLabel, ColumnLabel, CT)]
+
   // extend the given environment with names introduced by this FROM clause
   private[analyzer2] def extendEnvironment(base: Environment[MT]): Either[AddScopeError, Environment[MT]]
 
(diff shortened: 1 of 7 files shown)


saraid (approved):
You said something about SEARCH being a problem, but I'm not seeing it, so I'm approving.

soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/SoQLAnalyzerTest.scala
```diff
@@ -758,6 +758,33 @@ class SoQLAnalyzerTest extends FunSuite with MustMatchers with TestHelper {
...
+
+  test("Hidden columns - union") {
+    val tf1 = tableFinder(
+      (0, "aaaa-aaaa") -> D("n1" -> TestNumber, "n2" -> TestNumber).withHiddenColumns("n1"),
+      (0, "bbbb-bbbb") -> D("n3" -> TestNumber),
+      (0, "cccc-cccc") -> Q(0, "aaaa-aaaa", "select * union select n3 from @bbbb-bbbb"),
+    )
+    val analysis1 = analyze(tf1, "cccc-cccc", "select :*, *")
+
+    println(analysis1.statement.debugDoc)
```
saraid, line 785: Debug?

rjmac: Yep, oops.  This entire class's change was not meant to be checked in.

saraid (approved):
I see no problems.

PR #307, opened 2023-04-13
overload reduce_precision to always return the multitype of the input

PeterAustinMoore:
- ST_ReducePrecision can return the multi-version of a geometry, so we have to overload the function to ensure the return type is multitype.

### DIFF
--- a/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
+++ b/soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
@@ -240,9 +240,35 @@ object SoQLFunctions {
   val GeoCollectionExtractMultiPointFromPoint = mf("geo_multi_collection_mpt_mpt", FunctionName("geo_collection_extract"), Seq(SoQLMultiPoint), Seq.empty, SoQLMultiPoint)(
     NoDocs
   )
-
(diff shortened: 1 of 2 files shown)


soql-stdlib/src/main/scala/com/socrata/soql/functions/SoQLFunctions.scala
```diff
@@ -240,12 +240,18 @@ object SoQLFunctions {
...
   )
-
-  val ReducePrecision = f("reduce_precision", FunctionName("reduce_precision"), Map("a" -> GeospatialLike, "b" -> NumLike),
+  // Reducing precision _can_ return multipolygons from polygons, so we have to split out the polygon type
+  val ReducePrecision = f("reducePrecision", FunctionName("reduce_precision"), Map("a" -> (GeospatialLike - SoQLPolygon), "b" -> NumLike),
     Seq(VariableType("a"), VariableType("b")), Seq.empty, VariableType("a")) (
       "Reduce the precision of a given geometry, for example reduce_precision(to_point('POINT (1.234 10.675)'), 0.1) => POINT (1.2 10.6)",
       Example("Reduce to tens place", "SELECT reduce_precision(to_point('POINT (1.234 10.675)'), 0.1)", "")
     )
+  val ReducePolyPrecision = f("reducePolyPrecision", FunctionName("reduce_precision"), Map("a" -> CovariantSet(SoQLPolygon), "b" -> NumLike),
```
rjmac, line 249: Get rid of the "a"here, it's no longer necessary.

rjmac (approved):

PR #312, opened 2023-05-30
Synthetic column marker

rjmac:


saraid (approved):
Yeah, that all tracks.

PR #313, opened 2023-05-30
Parameter context for the new-analyzer

rjmac:


soql-stdlib/src/main/scala/com/socrata/soql/stdlib/analyzer2/Context.scala
```diff
@@ -0,0 +1,15 @@
+package com.socrata.soql.stdlib.analyzer2
+
+import com.rojoma.json.v3.util.{AllowMissing, AutomaticJsonCodec}
+
+@AutomaticJsonCodec
+case class Context(
+  @AllowMissing("Map.empty")
+  system: Map[String, String],
```
rlvoyer, line 8: I'm curious what examples of this system context map might look like. Would it make sense to write some tests that illustrate how this `Context` should be used?

soql-stdlib/src/main/scala/com/socrata/soql/stdlib/analyzer2/UserParameters.scala
```diff
@@ -0,0 +1,107 @@
...
+      .branch[FixedTimestamp]("fixed_timestamp")
+      .branch[FloatingTimestamp]("floating_timestamp")
+      .build
+  }
+}
+
+@AutomaticJsonCodec
+case class UserParameters(
+  @AllowMissing("Map.empty")
+  qualified: Map[CanonicalName, Map[HoleName, UserParameters.Value]],
```
rlvoyer, line 95: Can you remind via an example what difference is between qualified and unqualified `UserParameters`?

soql-stdlib/src/main/scala/com/socrata/soql/stdlib/analyzer2/Context.scala
rjmac: It's not really up to soql-reference to decide how it's used :)  - what this is for is so that Core can pass down info (specifically, user identity info) into the seoncdary so it can be used to support the secured views stuff.

But at this level, it's literally just a key-value map with no attached semantics beyond that.

rlvoyer (approved):
:shipit:

PR #318, opened 2023-06-15
EN-60580: Change how function docs are represented

rjmac:
* It's now a single aggregate object instead of a pair of fields
* This aggregate object also has a normal/deprecated/hidden status field
* Fix the SoQLDocs subproject (and make sure it's actually built in a normal build to prevent bitrot)

This will probably be a mildly disruptive change, though _most_ things neither build nor destructure Functions so it shouldn't be too bad.


saraid (approved):
:good:

PR #325, opened 2023-07-18
EN-60584: Mark various functions deprecated or hidden

rjmac:
Each such comment is commented with a reason for it to be in this state.


saraid (approved):
Comments good!

PR #326, opened 2023-07-31
Make the preserve ordering rewrite produce less redundant clauses

rjmac:
In the case where a query-chain orders one way and then orders by the same column another way.


saraid (approved):
Can't think of any problems with this.

PR #327, opened 2023-07-31
In TableFinder, allow setting null first/last on dataset ordering

rjmac:


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/MockTableFinder.scala
```diff
@@ -12,20 +12,32 @@ import com.socrata.soql.parsing.standalone_exceptions.LexerParserException
 import com.socrata.soql.parsing.StandaloneParser
 import com.socrata.soql.analyzer2.{TableFinder, DatabaseTableName, DatabaseColumnName, TableDescription, CanonicalName, TableMap, ParserUtil, FoundTables, UnparsedFoundTables, UnparsedTableMap, ScopedResourceName, MetaTypes, MetaTypeHelper}
 
+sealed abstract class OptionalBoolean {
+  def orElse(b: Boolean): Boolean
+}
+object OptionalBoolean {
+  implicit class Provided(value: Boolean) extends OptionalBoolean {
+    def orElse(b: Boolean) = value
+  }
```
rjmac, line 21: Scala hacks.  This lets the `withOrdering` method below have the usual null-positioning semantics ("null is greater than any other value") even though Scala doesn't allow optional parameters to depend on the values of other optional parameters in the same parameter-list.

saraid (approved):

PR #331, opened 2023-09-01
Make it possible to use the results of one iso check to inform another

rjmac:


saraid (approved):
Seems straightforward.

PR #342, opened 2023-09-26
Make "vertical slice" match more candidates

rjmac:
By making an overconstrained ORDER BY on a candidate be accepted when doing that match


saraid (approved):
I think that comparator is going in the correct direction, so, cool.

PR #353, opened 2023-10-20
Add make_point

rjmac:
Also add a thing to make the 2.12 compiler stop whining that I'm using higher-kinded types without warning it first.


rlvoyer (approved):
:shipit:

PR #354, opened 2023-10-23
Make the error change less churn-inducing

rjmac:
* Make `"position"` an optional top-level property of the encoded error
* `"code"` ⇒ `"type"` as the for the error type tag

Plus some QOL improvements:

* `errorCodecs` by convention now defaults to an empty builder parameter
* `SoQLErrorCodec#result` is no longer overloaded (though its arguments are)

And a correction:
* No longer expose ErrorHierarchyCodec stuff

### DIFF
--- a/soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/SoQLErrorTest.scala
+++ b/soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/SoQLErrorTest.scala
@@ -10,7 +10,7 @@ import com.socrata.soql.analyzer2.mocktablefinder._
 import com.socrata.soql.util.{SoQLErrorCodec, EncodedError}
 
 class SoQLErrorTest extends FunSuite with MustMatchers with TestHelper {
-  lazy val codecs = SoQLError.errorCodecs[Int, SoQLError[Int]](new SoQLErrorCodec.ErrorCodecs)
+  lazy val codecs = SoQLError.errorCodecs[Int, SoQLError[Int]]()
   lazy val jsonCodecs = codecs.build
 
   sealed trait OtherErrors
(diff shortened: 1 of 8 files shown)


soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/SoQLErrorTest.scala
```diff
@@ -51,6 +51,8 @@ class SoQLErrorTest extends FunSuite with MustMatchers with TestHelper {
     val Left(err@ParserError.ExpectedToken(_, _, _, _)) = tf.findTables(0, rn("q"))
     val encoded = jsonCodecs.encode(err)
 
+    println(encoded)
+
```
rjmac, line 55: oops, didn't mean to leave this in!

soql-utils/src/main/scala/com/socrata/soql/util/EncodedError.scala
```diff
@@ -13,10 +15,11 @@ import com.socrata.soql.environment.ScopedResourceName
 // We want our user-actionable error codes to have a common format.  This format will be:
 //
 // {
-//   "code": "some-identifier-for-machines",
+//   "type": "some-identifier-for-machines",
```
saraid, line 18: "message" -> "english" I'm fine with.
"code" -> "type" is ... fine, but also why?

saraid (approved):

soql-utils/src/main/scala/com/socrata/soql/util/EncodedError.scala
rjmac: Because that's what it was before, and there's code (in dsmapi in particular) that expects to find these names.  Rather than making it willing to accept either, it's nicer to make there be only one format.

PR #355, opened 2023-10-24
Window frame rework

rjmac:
* Function and Monomorphic function are typed better (they now have a single FunctionType parameter which describes the kind of function, rather than a pair of booleans where (true, true) was meaningless)
* Frame is non-optional on WindowedFunctionCall
* On some window functions, frame is meaningless.  Note this in the FunctionType, but do not reject SQL that provides it.


zachlefevre (approved):

zachlefevre (approved):

saraid (approved):
Can't see any problems.

PR #365, opened 2023-12-27
The Great Position Rework

rjmac:
The major point of this change is to be able to properly distinguish synthetic positions (that do not correspond to any user-provided soql) from non-synthetic ones.  The minor one is to change the naming of "physical" and "logical" positions to more clearly specify what they're actually for - in particular, the latter is used when reporting an error where there's no problem with the expression per se, but the expression cannot be used for whatever reason where the user is attempting to use it.  These can be different if an expression is a named select column.

One side effect of this is that scopedresourcename and position are now bundled together into a new type Source.


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/PositionInfo.scala
```diff
@@ -2,140 +2,200 @@ package com.socrata.soql.analyzer2
...
-          new AtomicPositionInfo(
-            logicalSource = buffer.read[Option[ScopedResourceName[RNS]]](),
-            logicalPosition = buffer.read[Position](),
-            physicalSource = buffer.read[Option[ScopedResourceName[RNS]]](),
-            physicalPosition = buffer.read[Position]()
-          )
-        case other =>
-          fail("Unknown AtomicPositionInfo tag " + other)
+      buffer.version match {
+        case Version.V0 =>
```
rjmac, line 72: Since we've got internal users now, it's time to start actually using the serialization-versioning stuff to make sure deploys don't cause query outages (the error format has changed too, but those happen way less frequently and so will hopefully not actually be a problem during the brief period of inconsistency during a deploy)

soql-environment/src/main/scala/com/socrata/soql/environment/Source.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.util.parsing.input.{Position, NoPosition}
+
+import com.rojoma.json.v3.codec.{JsonEncode, JsonDecode}
+import com.rojoma.json.v3.util.{SimpleHierarchyEncodeBuilder, SimpleHierarchyDecodeBuilder, InternalTag, AutomaticJsonEncodeBuilder, AutomaticJsonDecodeBuilder}
+
+import com.socrata.soql.jsonutils.HierarchyImplicits._
+import com.socrata.soql.jsonutils.PositionCodec
+
+sealed abstract class Source[+RNS] {
+  val scopedResourceName: Option[ScopedResourceName[RNS]]
```
rlvoyer, line 12: Is `RNS` short for `ResourceNameScope`?

soql-environment/src/main/scala/com/socrata/soql/environment/Source.scala
```diff
@@ -0,0 +1,58 @@
...
+  val position: Position
+  def withPosition(pos: Position): Source[RNS]
+  def mapSRN[RNS2](f: ScopedResourceName[RNS] => ScopedResourceName[RNS2]): Source[RNS2]
+}
+
+object Source {
+  case class Anonymous(position: Position) extends Source[Nothing] {
```
rlvoyer, line 19: Can you put a short scala doc comment in here to help clarify how these `Source` types differ and how they're intended to be used?

rlvoyer (changes requested):
Some docstrings inline would be really useful for understanding their different use-cases. Otherwise, this looks sane to me.

soql-environment/src/main/scala/com/socrata/soql/environment/Source.scala
rjmac: Yep; it's an abbreviation used pervasively in the code.

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/SoQLError.scala
```diff
@@ -100,14 +128,15 @@ object TableFinderError {
...
-          position <- position(v)
         } yield {
-          PermissionDenied(source, position, fields.name)
+          PermissionDenied(source, fields.name)
         }
     }
   }
 
-  case class RecursiveQuery[+RNS](source: Option[ScopedResourceName[RNS]], position: Position, stack: Seq[CanonicalName]) extends TableFinderError[RNS]
+  case class RecursiveQuery[+RNS](definiteSource: Source[RNS], stack: Seq[CanonicalName]) extends TableFinderLookupError[RNS] {
```
saraid, line 137: Hm. Makes me wonder if the `CallerStack` concept should be generalized so it can be used here, too. 🤷

rjmac: Definitely not; this is an error, and its responsibility is saying "you had a recursive  query, the chain was a -> b -> c -> b oops (or whatever)" so all it needs to know is that sequence of canonical-names; the whole CallerStack is just that exact list but with each node annotated with some extra stuff the tablefinder cares about.

rjmac: test

saraid (approved):
Not seeing any issues.

PR #366, opened 2023-12-27
Column hints

rjmac:


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveTrivialSelects.scala
```diff
@@ -19,6 +19,9 @@ class RemoveTrivialSelects[MT <: MetaTypes] private () extends StatementUniverse
           var exprs = OrderedMap[AutoColumnLabel, Column]()
 
           for((label, namedExpr) <- selectList) {
+            if(namedExpr.hint != None) { // if a selected column has a hint, it's not trivial
+              return None
```
zachlefevre, line 23: sure makes me miss `thing?` in Rust.

rjmac: Yes.

zachlefevre:
What is a hint actually _for_ though?

rjmac:
> What is a hint actually _for_ though?

Solving the core-column-formatting problem where core expects that soql which doesn't actually transform a value to also preserve that value's formatting when it's rendered.   This lets core attach whatever metadata it needs to columns and then retrieve it from the query result's schema to figure out what formatting should apply.  It is alas a very niche usecase and very core-specific, but not _so_ core-specific that anything but core needs to know anything but "there's some opaque data attached to this column".

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveTrivialSelects.scala
```diff
@@ -19,6 +19,9 @@ class RemoveTrivialSelects[MT <: MetaTypes] private () extends StatementUniverse
           var exprs = OrderedMap[AutoColumnLabel, Column]()
 
           for((label, namedExpr) <- selectList) {
+            if(namedExpr.hint != None) { // if a selected column has a hint, it's not trivial
```
saraid, line 22: Hm. This feels like it'll make for very few cases of trivial selects. Not sure how I feel about that.

rjmac: Hm.

saraid (approved):
Not finding any concerns, so, 👍

saraid (approved):
How often do I flag something as odd, decide it's not a big deal, and then find out it was actually a bug?

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/Merger.scala
```diff
@@ -135,89 +135,124 @@ class Merger[MT <: MetaTypes](and: MonomorphicFunction[MT#ColumnType]) extends S
...
+              orderBy = mergeOrderBy(aLabel, a.selectedExprs, aOrder, bOrder),
+              limit = bLim,
+              offset = bOff,
+              search = None,
+              hint = aHint ++ bHint)
+          ).filter(_.isAggregated) // again, just in case the aggregation was implicit and our merge removed it
+        case _ =>
+          debug("decline to merge - unknown pattern")
+          None
+      }
```
rjmac, line 221: This chunk is just a reindent.

saraid (approved):

PR #368, opened 2024-01-03
Outer joins can make primary key columns non-primary

rjmac:


saraid (approved):
Probably worth making a note _somewhere_ that "uniqueness" means "unique-and-not-null", but eh.

PR #383, opened 2024-03-19
Remove trivial joins and from clauses

rjmac:


soql-toy/src/main/scala/com/socrata/soql/AnalyzeFoundTables.scala
saraid, line 1: This is fine and all, but seems like a random thing to include in this PR.

saraid (approved):

soql-toy/src/main/scala/com/socrata/soql/AnalyzeFoundTables.scala
rjmac: Oh!  This wasn't supposed to be included, it's some debugging code I had hanging around.

PR #385, opened 2024-04-04
Fix an edge-case involving rollups and count(*)

rjmac:
See the new test in RollupExactTest for details

### DIFF
--- a/soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/rollup/RollupExactTest.scala
+++ b/soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/rollup/RollupExactTest.scala
@@ -379,4 +379,28 @@ class RollupExactTest extends FunSuite with MustMatchers with RollupTestHelper w
       TestRollupExact(select, rollup, analysis.labelProvider) must be (None)
     }
   }
+
+  test("A parameterless aggregate cannot be used with a rollup that has the same grouping which does not produce that aggregate as output") {
+    val tf = tableFinder(
+      (0, "twocol") -> D("text" -> TestText, "num" -> TestNumber),
+    )
+
+    val rollup = TestRollupInfo(1, "rollup", tf, "select text, sum(num) from @twocol group by text")
+
+    val Right(foundTables) = tf.findTables(0, "select text, count(*) from @twocol group by text", Map.empty)
+    val Right(analysis) = analyzer(foundTables, UserParameters.empty)
+    val select = analysis.statement.asInstanceOf[Select]
+
+    // So the bug was that was incorrectly saying "I _can_ use this
+    // rollup to answer this query (and generating the nonsense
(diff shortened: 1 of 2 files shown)


soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/rollup/RollupTestHelper.scala
```diff
@@ -122,7 +122,7 @@ trait RollupTestHelper extends TestHelper { this: Assertions =>
   object TestRollupInfo {
     def apply(id: Int, name: String, tf: TableFinder[TestMT], soql: String): TestRollupInfo = {
       val Right(foundTables) = tf.findTables(0, soql, Map.empty)
-      val analysis = analyzer(foundTables, UserParameters.empty) match {
+      val analysis = systemColumnPreservingAnalyzer(foundTables, UserParameters.empty) match {
```
rjmac, line 125: Oops, didn't mean to check this line in; it's not necessary for the repro.

soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/rollup/RollupExactTest.scala
```diff
@@ -379,4 +379,28 @@ class RollupExactTest extends FunSuite with MustMatchers with RollupTestHelper w
       TestRollupExact(select, rollup, analysis.labelProvider) must be (None)
     }
   }
+
+  test("parameterless aggregate one-to-one rewrite bug") {
```
rlvoyer, line 383: Nit: I think it would helpful as a reader of this code if the spec description actually called out desired behavior rather than referring to the bug.

Something like this?
```scala
test("rollup on an an aggregate with a parameter cannot be applied to a query with the same groupby clause but a different aggregate and parameter")
```

rjmac: That's not quite the bug though; it was specifically the fact that "count(*)" has no parameters that was the trigger.  It is indeed not an ideal name though.

dnabilali (approved):

soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/rollup/RollupExactTest.scala
```diff
@@ -379,4 +379,28 @@ class RollupExactTest extends FunSuite with MustMatchers with RollupTestHelper w
       TestRollupExact(select, rollup, analysis.labelProvider) must be (None)
     }
   }
+
+  test("A parameterless aggregate cannot be used with a rollup that has the same grouping which does not produce that aggregate as output") {
```
rlvoyer, line 383: 👍

saraid (approved):
I think it makes sense to me.

PR #387, opened 2024-04-10
Add a "preserve ordering with output columns" rewrite pass

rjmac:


saraid (approved):
It's weird that PreserveOrdering notes the caller might not get the output columns even when asking for them, but I'm not spotting a case where they'd not get returned.

rjmac:
> It's weird that PreserveOrdering notes the caller might not get the output columns even when asking for them, but I'm not spotting a case where they'd not get returned.

Yeah, when I wrote that I was thinking about `DISTINCT`, but `DISTINCT` already requires that orderings are selected.

PR #393, opened 2024-05-10
EN-66000: Infrastructure for generalized convert-to-text

rjmac:
For use in places where values get (sometimes implicitly) converted to text.

### DIFF
--- a/soql-types/src/main/scala/com/socrata/soql/types/obfuscation/CryptProvider.scala
+++ b/soql-types/src/main/scala/com/socrata/soql/types/obfuscation/CryptProvider.scala
@@ -5,8 +5,9 @@ import java.security.SecureRandom
 import org.bouncycastle.crypto.params.KeyParameter
 import org.bouncycastle.crypto.engines.BlowfishEngine
 
-class CryptProvider(key: Array[Byte]) {
-  private val keyParam = new KeyParameter(key)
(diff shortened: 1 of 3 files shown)


soql-types/src/main/scala/com/socrata/soql/types/obfuscation/CryptProvider.scala
```diff
@@ -5,8 +5,9 @@ import java.security.SecureRandom
 import org.bouncycastle.crypto.params.KeyParameter
 import org.bouncycastle.crypto.engines.BlowfishEngine
 
-class CryptProvider(key: Array[Byte]) {
-  private val keyParam = new KeyParameter(key)
+class CryptProvider(keyMaterial: Array[Byte]) {
+  def key = java.util.Arrays.copyOf(keyMaterial, keyMaterial.length)
```
zachlefevre, line 9: Why are you doing the this copy-ing?

zachlefevre (approved):

soql-types/src/main/scala/com/socrata/soql/types/obfuscation/CryptProvider.scala
rjmac: Because Java arrays are mutable.  It's just a defensive copy (and it's only 72 bytes, so not very expensive at all)

PR #394, opened 2024-05-21
More functions!

rjmac:
* Field-syntax access for parts of floating_timestamps, dates, and times
** including fieldlike access to "date" and "time" parts of timestamps
* Add a time to a date to get a floating_timestamp
* Add or subtract an interval to a time to get a new time
* Subtract two times to get an interval
* Add an interval to a date to get a floating_timestamp (not sure why
  the discrepancy, but it's what PG does)
* Subtract an interval from a date to get a floating_timestamp
* Find the difference between two dates in days
* Truncate a date at year or month
* Convert timestamps, dates, times, intervals to ISO8601 via cast-to-text
* Convert row identifiers and row versions to text via cast

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rollup/RollupExact.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rollup/RollupExact.scala
@@ -511,11 +511,15 @@ class RollupExact[MT <: MetaTypes](
     }
 
     val groupBySubset =
-      select.groupBy.forall { sGroupBy =>
-        candidate.groupBy.exists { cGroupBy =>
-          sGroupBy.isIsomorphic(cGroupBy, rewriteInTerms.isoState)
-        } || candidate.groupBy.exists { cGroupBy =>
-          functionSubset(sGroupBy, cGroupBy, rewriteInTerms.isoState).isDefined
+      select.groupBy.iterator.map { e =>
+        e +: adHocRewriter(e)
+      }.forall { sGroupByPossibilities =>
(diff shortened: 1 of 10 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rollup/RollupExact.scala
```diff
@@ -830,9 +830,10 @@ class RollupExact[MT <: MetaTypes](
         case Some((selectedColumn, NamedExpr(_ : WindowedFunctionCall, _name, _hint, _isSynthetic), _)) if rollupContext.isCoarseningGroup =>
           log.debug("can't rewrite, windowed function call")
           None
-        case Some((selectedColumn, ne, functionExtract)) =>
-          assert(ne.expr.typ == sExpr.typ)
-          Some(functionExtract(PhysicalColumn[MT](newFrom.label, newFrom.tableName, columnLabelMap(selectedColumn), sExpr.typ)(sExpr.position.asAtomic)))
+        case Some((selectedColumn, rollupExpr, functionExtract)) =>
+          val result = functionExtract(PhysicalColumn[MT](newFrom.label, newFrom.tableName, columnLabelMap(selectedColumn), rollupExpr.typ)(sExpr.position.asAtomic))
+          assert(result.typ == sExpr.typ)
+          Some(result)
```
rjmac, line 836: This wasn't quite asserting on the right thing; before "functionExtract" never changed the type of the thing it was extracting from.  Now it can, so we assert that the result of the extraction is the same as the type of the source expr.

zachlefevre (approved):

PR #395, opened 2024-05-29
Extract building the alias analysis context into a public static func…

rjmac:
…tion

Because it's sometimes useful when doing a partial analysis


rjmac:
Actually no, this isn't useful, because `From` is itself post-analysis.

dnabilali (approved):

PR #396, opened 2024-05-29
Make hidden columns work with top-level table-op nodes

rjmac:


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/SoQLAnalyzer.scala
```diff
@@ -466,38 +466,63 @@ class SoQLAnalyzer[MT <: MetaTypes] private (
...
+    def analyzeTableOp(ctx: Ctx, op: TrueOp[ast.Select], from0: ImplicitFrom): FromStatement = {
+      val subCtx =
+        Ctx(
+          scope = ctx.scope,
+          scopedResourceName = None,
+          canonicalName = None,
+          primaryTableName = ctx.primaryTableName,
+          enclosingEnv = ctx.enclosingEnv,
+          udfParams = ctx.udfParams,
+          hiddenColumns = Set.empty,
```
rjmac, line 482: The diff doesn't highlight this change very well, but this line was altered (we're doing hidden-columns as a postprocessing step that removes them now, so as far as subqueries are concerned there are none)

zachlefevre (approved):

PR #397, opened 2024-05-30
Helper for looking up found tables by their database table name

rjmac:
It's possible (albeit uncommon) for a table to have more than one referencing ScopedResourceName so this produces a map from DTN to a non-empty seq of scoped resource names.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/TableMap.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/TableMap.scala
@@ -37,6 +37,24 @@ class TableMap[MT <: MetaTypes] private[analyzer2] (private val underlying: Map[
 
   type Self[MT <: MetaTypes] = TableMap[MT]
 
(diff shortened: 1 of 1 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/TableMap.scala
```diff
@@ -37,6 +37,24 @@ class TableMap[MT <: MetaTypes] private[analyzer2] (private val underlying: Map[
...
+              case None => NonEmptySeq(srn, Nil)
+            }
+            acc + (ds.name -> prepended)
+          case _ =>
+            acc
+        }
+      }
+    }
+  }
+
```
rjmac, line 57: The key missing piece to understanding this is that `underlying` is a `Map[Scope, Map[ResourceName, TableDescription]`, so we're accumulating the result over the outer map (via the first `foldLeft`) by accumulating over the result at each inner map (via the second `foldLeft`)

dnabilali (approved):

PR #399, opened 2024-06-26
Remove trivial subselect introduced by RemoveUnusedOrderBy

rjmac:
"Trivial" here means "selects some untransformed columns from a physical table, with no other processing".  The main use case is where it deletes a default ordering from a joined-to dataset, so it doesn't need to be a subselect anymore.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveUnusedOrderBy.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveUnusedOrderBy.scala
@@ -2,8 +2,13 @@ package com.socrata.soql.analyzer2.rewrite
 
 import com.socrata.soql.analyzer2
 import com.socrata.soql.analyzer2._
+import com.socrata.soql.collection._
 
 class RemoveUnusedOrderBy[MT <: MetaTypes] private () extends StatementUniverse[MT] {
+  type RelabelMap = Map[(AutoTableLabel, AutoColumnLabel), (DatabaseColumnName, DatabaseTableName)]
+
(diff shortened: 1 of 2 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveUnusedOrderBy.scala
```diff
@@ -2,8 +2,13 @@ package com.socrata.soql.analyzer2.rewrite
 
 import com.socrata.soql.analyzer2
 import com.socrata.soql.analyzer2._
+import com.socrata.soql.collection._
 
 class RemoveUnusedOrderBy[MT <: MetaTypes] private () extends StatementUniverse[MT] {
+  type RelabelMap = Map[(AutoTableLabel, AutoColumnLabel), (DatabaseColumnName, DatabaseTableName)]
```
saraid, line 8: Randomly, why `table, column -> column, table`? I guess it doesn't really matter, unless it does?

saraid (approved):

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveUnusedOrderBy.scala
rjmac: It doesn't really matter, no.  The DTN in a PhysicalColumn is just a random bit of denormalization that I've been not super happy with since I introduced it, and so returning it here I'm thinking of in terms of "additional data to provided to the PC constructor" rather than "part of the column identifier" (before starting to write the Relabeller, this was just `Map[(ATL, ACL), DCN]`)

PR #403, opened 2024-07-17
Add a "remove synthetic columns" rewrite pass

rjmac:
This is nicer than doing it as a postprocessing pass in core because it allows the removal to be propagated through the whole query via the "remove unused columns" pass.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
@@ -1,0 +1,90 @@
+package com.socrata.soql.analyzer2.rewrite
+
+import com.socrata.soql.analyzer2
+import com.socrata.soql.analyzer2._
+import com.socrata.soql.collection._
+
+class RemoveSyntheticColumns[MT <: MetaTypes] private (labelProvider: LabelProvider) extends StatementUniverse[MT] {
+  def rewriteStatement(stmt: Statement): Statement =
+    stmt match {
+      case ct@CombinedTables(_op, _left, _right) =>
+        if(ct.schema.values.exists(_.isSynthetic)) {
+          // there shouldn't ever actually be any synthetic columns to
+          // remove here, so this is "just in case".  We'll add a new
+          // wrapper select that drops the columns, rather than trying
+          // to rewrite left and right in an equivalent way.  Note we
+          // don't care about preserving our own labels, as this is
+          // the output schema for the analysis.
+
+          val newFrom = FromStatement(ct, labelProvider.tableLabel(), None, None)
+          Select(
+            Distinctiveness.Indistinct(),
+            OrderedMap() ++ ct.schema.iterator.flatMap { case (k, v) =>
+              if(v.isSynthetic) {
+                None
+              } else {
+                Some(labelProvider.columnLabel() -> NamedExpr(VirtualColumn[MT](newFrom.label, k, v.typ)(AtomicPositionInfo.Synthetic), v.name, v.hint, isSynthetic = false))
+              }
+            },
+            newFrom,
+            None, Nil, None, Nil, None, None, None, Set.empty
+          )
+        } else {
+          ct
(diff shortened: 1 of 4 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
```diff
@@ -0,0 +1,90 @@
...
+          // remove here, so this is "just in case".  We'll add a new
+          // wrapper select that drops the columns, rather than trying
+          // to rewrite left and right in an equivalent way.  Note we
+          // don't care about preserving our own labels, as this is
+          // the output schema for the analysis.
+
+          val newFrom = FromStatement(ct, labelProvider.tableLabel(), None, None)
+          Select(
+            Distinctiveness.Indistinct(),
+            OrderedMap() ++ ct.schema.iterator.flatMap { case (k, v) =>
```
zachlefevre, line 22: why are you prepending an empty orderedmap?

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
```diff
@@ -0,0 +1,90 @@
...
+              if(v.isSynthetic) {
+                None
+              } else {
+                Some(labelProvider.columnLabel() -> NamedExpr(VirtualColumn[MT](newFrom.label, k, v.typ)(AtomicPositionInfo.Synthetic), v.name, v.hint, isSynthetic = false))
```
zachlefevre, line 26: is this empty columnLabel() what you mean when you say `we don't care about preserving our own labels`?

zachlefevre, line 26: Is the `AtomicPositionInfo.Synthetic` indicative of this column not having any *real* location in the underlying query since we're generating it here?

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
rjmac: Easiest way to build an orderedmap incrementally  without making an intermediate collection

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
```diff
@@ -0,0 +1,90 @@
...
+              }
+            },
+            newFrom,
+            None, Nil, None, Nil, None, None, None, Set.empty
```
zachlefevre, line 30: So many ways to say Nothing lol

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/RemoveSyntheticColumns.scala
rjmac: `columnLabel()` is a 0-arg function that returns a fresh label.  But yes - what I mean by that is "before we rewrite, our schema is, say, `label1 -> synthetic text, label2 -> num` and after we rewrite our schema might be `label45 -> num`"

rjmac: Yep.  It's a different but related meaning of "synthetic" than the one in "synthetic column".  This means specifically "this expr does not correspond to any input soql source code"

rjmac: (vs. synthetic column which means "this column does not correspond to any explicitly-selected column in the soql source code", which is a subtle but distinct meaning since it's tied to column identity rather than source position)

zachlefevre (approved):
I have some questions, but no suggestions

PR #411, opened 2024-10-25
EN-69963: Be less lazy in `squash`

rjmac:
Squash is used to say "here are a bunch of exprs of possibly different types; for each type, select the most preferable one.  It was producing a Stream unnecessarily which was building up into an enormous force event that overflowed the stack on long argument lists.


saraid (approved):
...wow.

PR #417, opened 2024-12-10
Get rid of obsolete types

rjmac:
These types no longer exist in core at all:

* Money
* Phone

These types do exist, but there are no columns of these types and no way save via API access to create column of them, and they never did anything useful anyway:

* Object
* Array


zachlefevre (approved):
we hardly knew ye SoQLMoney

PR #420, opened 2024-12-20
Allow customizing the soql2 toy's environment

rjmac:


soql-toy/soql2.json
```diff
@@ -0,0 +1,74 @@
+// A sample config file for the "soql2" toy
```
zachlefevre, line 1: Does Rojoma allow parsing json with comments removed?

zachlefevre (approved):

soql-toy/soql2.json
rjmac: Yep

PR #421, opened 2024-12-30
Report recursive-query errors at the highest level possible

rjmac:
Before, if you were tablefinding from query A that called B that called A, the table finder would report the error at B.  Now it runs up the stack to find the start of the loop and reports the error there.  This is better because "interior" errors might not get shown to the end-user (as they may not have access to interior saved queries), so this maximizes the possibility for the error to be actionable.


zachlefevre (approved):
I think this makes sense!

PR #424, opened 2025-02-03
Old-analyzer/new-analyzer compatibility around joined column aliases

rjmac:
In the OA, you weren't allowed to reference semi-explicitly aliased foreign columns in other parts of the query, more or less accidentally. The NA allowed it.  This change makes the NA behavior closer to the OA behavior here.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/SoQLError.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/SoQLError.scala
@@ -827,7 +827,7 @@ object SoQLAnalyzerError {
             .append(err.name)
             .append("'")
           if(err.qualifier.isEmpty) {
-            msg.append("  (Note: joined columns must be fully qualified)")
+            msg.append(" (Note: joined columns must be fully qualified)")
           }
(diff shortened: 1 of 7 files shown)


saraid (commented):
This makes sense to me.

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/SoQLError.scala
```diff
@@ -820,8 +820,17 @@ object SoQLAnalyzerError {
...
-        def encode(err: NoSuchColumn[RNS]) =
-          result(Fields(err.qualifier, err.name), s"No such column `${err.qualifier.fold("")("@" + _ + ".")}${err.name}'", err.source)
+        def encode(err: NoSuchColumn[RNS]) = {
+          val msg = new StringBuilder
+          msg.append("No such column `")
+            .append(err.qualifier.fold("")("@" + _ + "."))
+            .append(err.name)
+            .append("'")
+          if(err.qualifier.isEmpty) {
+            msg.append("  (Note: joined columns must be fully qualified)")
```
saraid, line 830: There's an extra space here.
```suggestion
            msg.append(" (Note: joined columns must be fully qualified)")
```
If we care.

PR #427, opened 2025-02-12
Remove explicit LATERAL on UDF calls if it's not required

rjmac:


saraid (approved):
"Don't add LATERAL unnecessarily" seems like a reasonable thing to not be a rewrite pass.

PR #441, opened 2025-07-31
EN-81002: OR equality comparisions with number cols onto SEARCH filters

rjmac:


saraid (approved):
Not spotting any problems...

PR #446, opened 2025-08-27
EN-79270: Remove bouncycastle dependency

rjmac:
We want to depend on a different version in other systems, and it was causing conflicts.  Since all we use it for is :id and :version obfuscation, just bring in enough code to handle that ourselves.

On the plus side, I get to fulfill a long-standing wish re: not converting things to byte arrays!


saraid (approved):
Starting to get worried about how often we're rolling our own crypto.

rjmac:
> Starting to get worried about how often we're rolling our own crypto.

If this were for anything other than :id and :version obfuscation I'd agree with you, but all I want here is a pseudorandom permutation of 64-bit ints that can be changed per dataset.  Blowfish was chosen entirely for its block size rather than its cryptographic properties.

PR #458, opened 2026-03-26
Add a column to the generated FromSingleRow selection

rjmac:
We already ensure there's at least one selected column in cases where we're just not selecting anything, but this case was missed.


saraid (approved):
Huh. Vaguely surprised we do this.

PR #462, opened 2026-06-09
EN-93632: Wrapping queries in the TableFinder and SoQLAnalyzer

rjmac:
Adds the ability for a tablefinder to wrap any found table (dataset, view, udf) in an auxiliary query that can process its output before being consumed downstream.  Views and UDFs' wrapped queries have access to parameters if they want them.

A lot of this is touching places where we convert Positions to Sources, in order to allow things that were never dealing with synthetic queries before to produce synthetic Sources.  Really, AST nodes should have Sources directly, but they don't because soda2 still exists.

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/UnparsedTableMap.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/UnparsedTableMap.scala
@@ -67,11 +67,10 @@ object UnparsedTableMap {
     new mocktablefinder.MockTableFinder[MT](
       OrderedMap() ++ self.underlying.iterator.flatMap { case (rns, resources) =>
         resources.iterator.map { case (rn, desc) =>
-          // TODO: attach the wrapping queries to the MTF!!!
           val thing =
             desc match {
-              case UnparsedTableDescription.Dataset(_name, canonicalName, schema, ordering, pks, wq) =>
-                val base =
+              case UnparsedTableDescription.Dataset(_name, canonicalName, schema, ordering, pks, wrappingQuery) =>
+                var dataset: mocktablefinder.D[MT#ResourceNameScope, MT#ColumnType] =
                   mocktablefinder.D(schema.valuesIterator.map { case TableDescription.DatasetColumnInfo(n, t, _, _) => n.name -> t }.toSeq : _*).
                     withHiddenColumns(schema.valuesIterator.filter(_.hidden).map(_.name.name).toSeq : _*).
                     withCanonicalName(canonicalName.name).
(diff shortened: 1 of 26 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/UnparsedTableMap.scala
```diff
@@ -72,9 +67,10 @@ object UnparsedTableMap {
     new mocktablefinder.MockTableFinder[MT](
       OrderedMap() ++ self.underlying.iterator.flatMap { case (rns, resources) =>
         resources.iterator.map { case (rn, desc) =>
+          // TODO: attach the wrapping queries to the MTF!!!
```
saraid: So, is this TODO for this PR...?

rjmac: Yep!  I forgot all about that.

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/Util.scala
```diff
@@ -84,4 +85,39 @@ private[analyzer2] object Util {
...
   }
+
+  implicit class OptionEitherExt[A, B](private val underlying: Option[Either[A, B]]) extends AnyVal {
+    def transposeOptionEither: Either[A, Option[B]] =
+      underlying match {
+        case Some(Left(a)) => Left(a)
+        case Some(Right(b)) => Right(Some(b))
+        case None => Right(None)
+      }
+  }
```
saraid, line 96: Is it bad that I'm very delighted you also had to create this construct of an Option[Either] that can become an Either[Option]? 🙃

rjmac: Nope!  It's useful (..which is why it's in the _rust_ standard library under the name "transpose" going both ways.  I thought "transpose" was a little _too_ opaque, though.  And I didn't actually end up using this much in the PR, but it's nice to have)

soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/PreserveOrdering.scala
```diff
@@ -9,7 +9,7 @@ import com.socrata.soql.collection._
 import com.socrata.soql.environment.ColumnName
 import com.socrata.soql.functions.MonomorphicFunction
 
-class PreserveOrdering[MT <: MetaTypes] private (provider: LabelProvider) extends StatementUniverse[MT] {
+class PreserveOrdering[MT <: MetaTypes] private (provider: LabelProvider, stop: FromStatement[MT] => Boolean) extends StatementUniverse[MT] {
```
saraid, line 12: Okay, I've been over a large chunk of the PR at this point and I still don't know:

Why do we need this `stop` function? When is it used?

saraid: ... and the _moment_ I post this, I realize that GitHub didn't load the SoQLAnalyzer large diff. 😮‍💨

rjmac: So the idea is that adding a wrapping query should not accidentally destroy any ordering on the wrapped query.   So, since we know nothing about that query, we use the PreserveOrdering rewrite pass to thread it through - but we don't want to preserve _all_ the ordering, we just want to go back until we hit the wrapped query or any other view which the wrapper query uses.  `stop` detects that boundary and prevents further recursion.

It's used at the place we call PreserveOrdering in SoQLAnalyzer.scala.

saraid (approved):
My brain has melted, but I think I followed all the individual pieces.

PR #465, opened 2026-06-12
Ensure wrapping queries do not affect column hints

rjmac:
This is a fairly major refector, because column hints previously conflated "no hint" and "inherit hint".

Also gets rid of serialization version 6; I thought this was going to need a new serialization rev, but it turned out not to.


saraid (approved):
LGTM.

PR #468, opened 2026-06-24
Less aggressive merging

rjmac:
When merging `select ..., f() as x, ... |> select ...`, if `x` is referenced more than once on the right hand side, refuse to duplicate the nontrivial expression and refuse the merge, _unless_ the duplicates are top-level `group by` or `order by` expressions and the original expression was seen at the top-level of a selection-list entry

### DIFF
--- a/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/Merger.scala
+++ b/soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/Merger.scala
@@ -200,7 +200,7 @@ class Merger[MT <: MetaTypes](and: MonomorphicFunction[MT#ColumnType]) extends S
       val newStatement: Option[Select] =
         (a, b) match {
           case (a, b) if definitelyRequiresSubselect(a, aLabel, b) =>
-            debug("declaining to merge - definitely requires subselect")
+            debug("declining to merge - definitely requires subselect")
             None
           case (a, b@Select(bDistinct, bSelect, _oldA, None, Nil, None, Nil, bLim, bOff, None, bHint)) if !b.isAggregated =>
             // Just projection change + possibly limit/offset and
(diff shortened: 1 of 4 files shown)


soql-analyzer/src/main/scala/com/socrata/soql/analyzer2/rewrite/Merger.scala
```diff
@@ -148,105 +190,110 @@ class Merger[MT <: MetaTypes](and: MonomorphicFunction[MT#ColumnType]) extends S
...
-          val hint = ColumnHint.fromAnalyzedOption(b.schema(label).hint)
-          label -> namedExpr.copy(hint = hint)
+  ): Option[Statement] =
+    try {
+      // If we decide to merge this, we're going to create some flavor of
+      //   select merged_projection from bRejoin(FromStatement(a.from)) ...)
+      val newStatement: Option[Select] =
+        (a, b) match {
+          case (a, b) if definitelyRequiresSubselect(a, aLabel, b) =>
+            debug("declaining to merge - definitely requires subselect")
```
saraid:
```suggestion
            debug("declining to merge - definitely requires subselect")
```

soql-analyzer/src/test/scala/com/socrata/soql/analyzer2/SoQLAnalysisTest.scala
```diff
@@ -264,11 +296,11 @@ select count(*) where num = 3
     )
 
     val analysis = analyze(tf, "twocol", """
-select text, count(num) group by text |> select * where count_num = 5
+select text, count(num) group by text |> select count_num = 5
```
saraid, line 314: Is this an intentional change? Seems like a weird thing to pipe into.

rjmac: It is.  What it's testing is that it's willing to rewrite that `count_num` on the right side into `op$=(count(num), 5)` (i.e., into a non-top-level context)

rjmac: (whereas before this change count(num) would have been duplicated and it would have refused to merge with this new logic)

saraid (approved):

saraid (approved):

PR #473, opened 2026-07-31
Handle type assertions in the analyzer

rjmac:
This makes them disappear completely from the analysis (which makes rollup passes able to treat a type-asserted expression as "simple", for things like inlining UDF parameters or creating select list references)


saraid (approved):
Straightforward.

